PR #2, opened 2015-05-26
Fixes @method docblock for plugin

gabrielsch:
It creates IDE warnings, there is a required parameter

### DIFF
--- a/src/Controller/AbstractController.php
+++ b/src/Controller/AbstractController.php
@@ -41,7 +41,7 @@ use Zend\Stdlib\ResponseInterface as Response;
  * @method \Zend\Mvc\Controller\Plugin\Redirect redirect()
  * @method \Zend\Mvc\Controller\Plugin\Url url()
  * @method \Zend\View\Model\ConsoleModel createConsoleNotFoundModel()
- * @method \Zend\View\Model\ViewModel createHttpNotFoundModel()
+ * @method \Zend\View\Model\ViewModel createHttpNotFoundModel(Response $response)
  */
 abstract class AbstractController implements
     Dispatchable,


Ocramius:
@gabrielsch merged, thx!

`master`: zendframework/zend-mvc@40efd6e8706b06d4c6c3b6abb3161f3d810c90d8
`develop`: zendframework/zend-mvc@af8003aa819e972f7930641738a37e83878a010d

Ocramius:
@weierophinney how do I backport this sort of fix?

weierophinney:
@Ocramius I've just created a "release-2.4" branch on the zf2 repo; you can patch against that. If you take the patch from this PR (https://patch-diff.githubusercontent.com/raw/zendframework/zend-mvc/pull/2.patch), you can edit the paths on it and use `git am` to apply it against zf2 (so authorship is retained). Make sure you indicate the origin of the patch in the commit.

Ocramius:
That seems like a very painful approach, but I guess we don't have many alternatives for the LTS

weierophinney:
@Ocramius oh, and one more thing... we need to create a release-2.4 branch on this one, too, as we know for certain there will be a 2.4 release with the patch as well, to mirror what goes in the framework. We'll need to be very careful about what we deem is a _critical_ fix, to minimize these backports.

weierophinney:
@Ocramius since this is just an IDE typehint, I don't think we need to backport it. It's not a critical functionality, just a convenience issue.

Ocramius:
@weierophinney do we really need a 2.4 branch in this one as well? Can't we require zendframework/zendframework:~2.4.0 instead? :lazy:

Ocramius:
^ obviously, I was kidding ^

PR #4, opened 2015-06-04
[ZF3] [BC-BREAK] Remove ServiceLocatorAwareInterface from AbstractController

stefanotorresi:
As discussed in zendframework/zf2#5168.

Mantaining BC is actually as easy as implementing `ServiceLocatorAwareInterface` explicitly and importing `ServiceLocatorAwareTrait` into userland controller classes.

This is just one little step towards zendframework/zf2#6068.

### DIFF
--- a/test/Controller/TestAsset/ServiceLocatorAwareController.php
+++ b/test/Controller/TestAsset/ServiceLocatorAwareController.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * @author Stefano Torresi (http://stefanotorresi.it)
+ * @license See the file LICENSE.txt for copying permission.
+ * ************************************************
+ */
+
+namespace ZendTest\Mvc\Controller\TestAsset;
+
+use Zend\ServiceManager\ServiceLocatorAwareInterface;
+use Zend\ServiceManager\ServiceLocatorAwareTrait;
+
+class ServiceLocatorAwareController extends SampleController implements ServiceLocatorAwareInterface
+{
+    use ServiceLocatorAwareTrait;
+}
--- a/src/Controller/AbstractController.php
+++ b/src/Controller/AbstractController.php
@@ -46,8 +46,7 @@ use Zend\Stdlib\ResponseInterface as Response;
 abstract class AbstractController implements
     Dispatchable,
     EventManagerAwareInterface,
-    InjectApplicationEventInterface,
-    ServiceLocatorAwareInterface
+    InjectApplicationEventInterface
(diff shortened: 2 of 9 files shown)


awartoft:
Woho :+1:

gianarb:
:+1:

weierophinney:
I hate to be constantly the voice of doom, but this "feature" is a no-starter from a migration stand-point.

Let's consider what a user would do on upgrading:
- Whoa! _Nothing_ works!
- What do I need to do?
- Oh, I have to go into **each** of my controllers and:
  - Import the ServiceLocatorAwareInterface
  - Import the ServiceLocatorAwareTrait
  - Add a `use ServiceLocatorAwareTrait` inside my class definition

All that, just to get the application back working the way it was.

**IF** you provide a tool that can script the above to aid in migration, I'll consider it, as I really, really hate that I capitulated and made controllers service-aware in the first place. But that tool is going to be necessary if we want folks to actually migrate from ZF2 to ZF3. And I really do, because the migration from ZF1 to ZF2 has not happened, precisely because it's very, very hard.

Ocramius:
PLING!
Suggestion: stop documenting the `AbstractController` and instead suggest a
no-inheritance approach for our users?

```
class MyController
{
    public function __construct($helpers) {
        $this->helpers = $helpers;
    }
    public function __invoke($request, $response, $next) {
        // LOGIC, DO YOU SPEAK IT?!
        return $next($request,
$response->withHeader('Location: ' . $this->helpers->url('homepage')));
    }
}
```

End of service announcement.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

stefanotorresi:
@Ocramius I suppose that's exactly what the new middleware approach will be about, but the existing MVC layer still needs some love.

@weierophinney Fair point; I'll take the burden to provide a migration tool (which will be full of notices of death). Anything to make this happen...

weierophinney:
@stefanotorresi Many thanks. :) Migration is a key concern; I'm quite happy with most of the changes, but anything that breaks current apps, I really, really want to provide automated migration tooling whenever possible.

test/Controller/TestAsset/ServiceLocatorAwareController.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
```
Maks3w, line 2: wrong file block

Maks3w:
Supersede by #36

PR #9, opened 2015-06-15
Hops not used because of unset in get method.

riznmage:
Calling getMessagesFromContainer() results in clearing the messages in the session, regarding the hops, defined when adding a message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ install:
 
 script:
   - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then ./vendor/bin/phpunit --coverage-clover clover.xml ; fi
-  - if [[ $EXECUTE_TEST_COVERALLS != 'true' ]]; then ./vendor/bin/phpunit --verbose ; fi
-  - ./vendor/bin/php-cs-fixer fix -v
+  - if [[ $EXECUTE_TEST_COVERALLS != 'true' ]]; then ./vendor/bin/phpunit ; fi
+  - if [[ $EXECUTE_CS_CHECK == 'true' ]]; then ./vendor/bin/php-cs-fixer fix -v --diff --dry-run ; fi
 
 after_script:
   - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then ./vendor/bin/coveralls ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
   - travis_retry composer install --no-interaction --ignore-platform-reqs
 
 script:
-  - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then ./vendor/bin/phpunit --coverage-clover clover.xml ; fi
+  - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then ./vendor/bin/phpunit --coverage-clover clover.xml --verbose ; fi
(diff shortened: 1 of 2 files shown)


weierophinney:
Please provide unit tests for this; right now I _still_ don't understand what the expectations are. Unit tests will clarify that.

Additionally, please look at the travis-ci output, and run `php-cs-fixer`; you have CS issues currently.

riznmage:
When hops are used, the key gets expired on the first request, no matter what number is used for the hops.

weierophinney:
@riznmage Please write _tests_ for that behavior, then.

Tests help us:
- understand the expectations properly.
- document the expectations for others.
- ensure that future changes do not _break_ those expectations.

riznmage:
The Travis CI build failed because of other tests, not related to the issue. You can check https://travis-ci.org/zendframework/zend-mvc/jobs/66870601

A test cannot be written, since a second request is needed to be made to test this. Basically, the expected behavior is this:
1. In a controller action, a message is set:
   `$this->flashMessenger()->addMessage("Message text", FlashMessenger::NAMESPACE_ERROR, 2);`
   This message uses 2 hops, which means that it will be kept for 2 more requests in the session.
2. A redirect is made from the same action:
   `return $this->redirect()->toRoute("someRouteName");`
   On the redirected page, the message is available.
3. Then a new request is made either by the user, or via AJAX/ESI. Judging by the value of the hops (2), the message should be available on this page, too, but it isn't.

Maks3w:
@riznmage Travis-CI it's failing because the PR does not follow the CS guidelines

https://travis-ci.org/zendframework/zend-mvc/jobs/66870601#L433

Execute `./vendor/bin/php-cs-fixer fix -v` in the project root

weierophinney:
> A test cannot be written, since a second request is needed to be made to test this.

You can mock the session manager in order to emulate multiple requests.

Maks3w:
@riznmage Your last commit is erroneus. You have to execute the command in your local machine and don't modify travis configuration which is out of the scope.

riznmage:
@weierophinney I haven't writter a unit test so far. I'm just reporting a bug. Read https://github.com/zendframework/zend-mvc/pull/9#issuecomment-112082949 thoroughly and you will get what I'm talking about.

weierophinney:
@riznmage Please review [my earlier comment](https://github.com/zendframework/zend-mvc/pull/9#issuecomment-112068654). I'm asking you to write a test so that we can understand the behavior expected _and_ ensure that once corrected it _stays_ corrected later.

If you need assistance writing the unit test, you can jump into the #zftalk.dev channel on Freenode IRC and ask.

riznmage:
> Execute ./vendor/bin/php-cs-fixer fix -v in the project root
> How exactly? @Maks3w

riznmage:
Okay.. it seems to be a mistake that I made. I was using `getMessages()` instead of `getCurrentMessages()`. Can you explain me what is the difference between these two? Why is `getMessages()` clearing the messages from the session container?

weierophinney:
`getMessages()` is the one to use when accessing messages set in the _previous_ request (which is the general reason for using FlashMessenger). `getCurrentMessages()` is the method to use when you want access to messages set during the _current_ request.

`getMessages()` will clear the container of any expired messages, so that the next invocation will only retrieve those that still have request hops available (which is what the `hops` setting is for); that's likely the behavior you are seeing.

PR #14, opened 2015-07-13
Added factory method in TranslatorAwareTreeRouteStack for missing option injection

agopaul:
Overrided the class factory method to import option 'text_domain' using the already present setter setTranslatorTextDomain().

This is necessary to be able to instantiate the TranslatorAwareTreeRouteStack router from config with the option:

```
<?php
return array(
    'router' => array(
        'router_class' => 'Zend\Mvc\Router\Http\TranslatorAwareTreeRouteStack',
        'text_domain' => 'router',
        'routes' => array(...),
    ),

    ///
);
```

Both tests and php-cs-fixer ran ok.

### DIFF
--- a/src/Router/Http/TranslatorAwareTreeRouteStack.php
+++ b/src/Router/Http/TranslatorAwareTreeRouteStack.php
@@ -47,7 +47,7 @@ class TranslatorAwareTreeRouteStack extends TreeRouteStack implements Translator
      * @param  array|Traversable $options
      * @return TranslatorAwareTreeRouteStack
      */
-    public static function factory($options = array())
+    public static function factory($options = [])
     {
         $instance = parent::factory($options);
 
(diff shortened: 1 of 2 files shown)


agopaul:
I'm still wrapping my head around the fact that php-cs-fixer doesn't give any warning on the brackets style in the local environment, but Travis does.
Well, now the build is ok

agopaul:
Nevermind, the PR is useless given that apparently the Translator is not automatically injected into TranslatorAwareTreeRouteStack, even if it implements TranslatorAwareInterface.

I'm not sure why this router implements TranslatorAwareInterface, given that the RouterFactory doesn't even inject the translator.

The only way to inject the translator and set the text domain is to do it in the module bootstrap, or maybe overriding the RouterFactory (without touching zf).

If I'll have the time to do it, I'll ask on the list or IRC and eventually open a PR for RouterFactory

PR #18, opened 2015-07-31
Fixes #11 : add null| to annotation to routeMatch

### LINKED ISSUE #11: `Zend\Mvc\MvcEvent#getRouteMatch()` should be annotated `RouteMatchInterface|null`

samsonasik:
Fixes #11


samsonasik:
@Ocramius is this change what you meant ?

PR #21, opened 2015-08-11
Unittests failing

weierophinney:
I just cloned this repo and unittests fails for both master and 2.5:

```
1) ZendTest\Mvc\Controller\Plugin\FilePostRedirectGetTest::testCorrectInputDataMerging
Failed asserting that false is true.

... zend-mvc/test/Controller/Plugin/FilePostRedirectGetTest.php:405
```


weierophinney:
This looks like a change in the zend-session library; @mwillbanks - can you look into this please, as I know you've been merging zend-session changes recently...

mwillbanks:
Reassigned back to @weierophinney as this was due to https://github.com/zendframework/zend-inputfilter/pull/7

weierophinney:
The issue is due to how the "required" validation failure message is injected after zendframework/zend-inputfilter#7; the patch I've introduced just updates the test to be more robust in how it finds that message in the returned validation failure message list.

PR #26, opened 2015-09-08
Fix 'Call to a member function hasChildren() on boolean'

skl:
- This seems to occur when the DefaultRenderingStrategy tries to render
  a result that is not an instance of Zend\View\Model\ModelInterface, in
  my case the result was a boolean true.
- I'm checking for ModelInterface rather than ConsoleModel as the render
  method seems to work with any ModelInterface, not just ConsoleModel

PR #27, opened 2015-09-08
Fix 'Call to a member function hasChildren() on boolean'

skl:
- This seems to occur when the DefaultRenderingStrategy tries to render
  a result that is not an instance of Zend\View\Model\ModelInterface, in
  my case the result was a boolean true.
- I'm checking for ModelInterface rather than ConsoleModel as the render
  method seems to work with any ModelInterface, not just ConsoleModel

Supersedes #26 due to rebase incorporating feedback.


skl:
Sorry looks like I should've forced-push to update #26 rather than delete and push which creates a new PR... next time.

weierophinney:
@skl In the future, you can push as many times as you want to the branch from which you created a PR; it will update the PR with new commits and the cumulative changes. When you force push, it simply replaces the commits and changes.

weierophinney:
Tagged in both release-2.5.2 and release-2.4.8

skl:
Thanks :smile:

PR #31, opened 2015-09-30
Refactored to use upcoming EventManager v3

weierophinney:
Updates all code to work with the upcoming v3 of EventManager. In particular:
- Updated the EventManager factory to inject the SharedEventManager at instantiation, if that service is present.
- Updated the EventManager initializer to inject not only if no EM instance is present, but also if the current EM instance has no shared manager present.  The latter is a good indication that the EM instance was lazy-loaded by the getter.
- Updated the Application constructor to pass arguments. This was necessary due to a strange circular dependency with the changes to the initializer. It's also better DI.
- Fixed all `attachAggregate()` calls to use `$aggregate->attach()`, and ensured there were no calls to `$events->attach()` that were attaching aggregates.
- Reviewed all `trigger()` calls for appropriate signatures, updating those that needed it. In many cases, an additional call to `$event->setName()` was also required.

This can likely be merged to develop sooner rather than later, as it represents a work-in-progress towards the updated MVC.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,7 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.6-dev",
-            "dev-develop": "2.7-dev"
+            "dev-develop": "3.0-dev"
         }
     },
     "autoload-dev": {
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,12 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-eventmanager": "~2.5",
+        "zendframework/zend-eventmanager": "dev-develop as 2.7.0",
         "zendframework/zend-servicemanager": "~2.5",
         "zendframework/zend-hydrator": "~1.0",
         "zendframework/zend-form": "~2.6",
-        "zendframework/zend-stdlib": "~2.7"
+        "zendframework/zend-stdlib": "~2.7",
+        "container-interop/container-interop": "^1.1"
     },
     "require-dev": {
         "zendframework/zend-authentication": "~2.5",
@@ -31,14 +32,14 @@
         "zendframework/zend-inputfilter": "~2.5",
         "zendframework/zend-json": "~2.5",
         "zendframework/zend-log": "~2.5",
-        "zendframework/zend-modulemanager": "~2.6",
+        "zendframework/zend-modulemanager": "dev-develop as 2.7.0",
         "zendframework/zend-session": "~2.5",
         "zendframework/zend-serializer": "~2.5",
         "zendframework/zend-text": "~2.5",
         "zendframework/zend-uri": "~2.5",
         "zendframework/zend-validator": "~2.5",
         "zendframework/zend-version": "~2.5",
-        "zendframework/zend-view": "~2.5",
+        "zendframework/zend-view": "dev-develop as 2.6.0",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },
--- a/src/Application.php
+++ b/src/Application.php
@@ -365,6 +365,7 @@ class Application implements
 
         $event->setName(MvcEvent::EVENT_FINISH);
         $events->triggerEvent($event);
+
         return $this;
     }
 }
--- a/src/Application.php
+++ b/src/Application.php
@@ -12,6 +12,7 @@ namespace Zend\Mvc;
 use Zend\EventManager\EventManagerAwareInterface;
 use Zend\EventManager\EventManagerInterface;
 use Zend\ServiceManager\ServiceManager;
+use Zend\Stdlib\RequestInterface;
 use Zend\Stdlib\ResponseInterface;
 
 /**
@@ -103,15 +104,18 @@ class Application implements
      * @param mixed $configuration
      * @param ServiceManager $serviceManager
      */
-    public function __construct($configuration, ServiceManager $serviceManager)
-    {
+    public function __construct(
+        $configuration,
+        ServiceManager $serviceManager,
+        EventManagerInterface $events,
+        RequestInterface $request,
+        ResponseInterface $response
+    ) {
         $this->configuration  = $configuration;
         $this->serviceManager = $serviceManager;
-
-        $this->setEventManager($serviceManager->get('EventManager'));
-
-        $this->request        = $serviceManager->get('Request');
-        $this->response       = $serviceManager->get('Response');
+        $this->setEventManager($events);
+        $this->request        = $request;
+        $this->response       = $response;
     }
 
     /**
@@ -142,19 +146,20 @@ class Application implements
         $listeners = array_unique(array_merge($this->defaultListeners, $listeners));
 
         foreach ($listeners as $listener) {
-            $events->attach($serviceManager->get($listener));
+            $serviceManager->get($listener)->attach($events);
         }
 
         // Setup MVC Event
         $this->event = $event  = new MvcEvent();
+        $event->setName(MvcEvent::EVENT_BOOTSTRAP);
         $event->setTarget($this);
-        $event->setApplication($this)
-              ->setRequest($this->request)
-              ->setResponse($this->response)
-              ->setRouter($serviceManager->get('Router'));
+        $event->setApplication($this);
+        $event->setRequest($this->request);
+        $event->setResponse($this->response);
+        $event->setRouter($serviceManager->get('Router'));
 
         // Trigger bootstrap events
-        $events->trigger(MvcEvent::EVENT_BOOTSTRAP, $event);
+        $events->triggerEvent($event);
         return $this;
     }
 
@@ -294,13 +299,15 @@ class Application implements
         };
 
         // Trigger route event
-        $result = $events->trigger(MvcEvent::EVENT_ROUTE, $event, $shortCircuit);
+        $event->setName(MvcEvent::EVENT_ROUTE);
+        $result = $events->triggerEventUntil($shortCircuit, $event);
         if ($result->stopped()) {
             $response = $result->last();
             if ($response instanceof ResponseInterface) {
+                $event->setName(MvcEvent::EVENT_FINISH);
                 $event->setTarget($this);
                 $event->setResponse($response);
-                $events->trigger(MvcEvent::EVENT_FINISH, $event);
+                $events->triggerEvent($event);
                 $this->response = $response;
                 return $this;
             }
@@ -311,14 +318,16 @@ class Application implements
         }
 
         // Trigger dispatch event
-        $result = $events->trigger(MvcEvent::EVENT_DISPATCH, $event, $shortCircuit);
+        $event->setName(MvcEvent::EVENT_DISPATCH);
+        $result = $events->triggerEventUntil($shortCircuit, $event);
 
         // Complete response
         $response = $result->last();
         if ($response instanceof ResponseInterface) {
+            $event->setName(MvcEvent::EVENT_FINISH);
             $event->setTarget($this);
             $event->setResponse($response);
-            $events->trigger(MvcEvent::EVENT_FINISH, $event);
+            $events->triggerEvent($event);
             $this->response = $response;
             return $this;
         }
@@ -350,8 +359,12 @@ class Application implements
     {
         $events = $this->events;
         $event->setTarget($this);
-        $events->trigger(MvcEvent::EVENT_RENDER, $event);
-        $events->trigger(MvcEvent::EVENT_FINISH, $event);
+
+        $event->setName(MvcEvent::EVENT_RENDER);
+        $events->triggerEvent($event);
+
+        $event->setName(MvcEvent::EVENT_FINISH);
+        $events->triggerEvent($event);
         return $this;
     }
 }
--- a/src/Controller/AbstractController.php
+++ b/src/Controller/AbstractController.php
@@ -109,13 +109,14 @@ abstract class AbstractController implements
         $this->response = $response;
 
         $e = $this->getEvent();
-        $e->setRequest($request)
-          ->setResponse($response)
-          ->setTarget($this);
+        $e->setName(MvcEvent::EVENT_DISPATCH);
+        $e->setRequest($request);
+        $e->setResponse($response);
+        $e->setTarget($this);
 
-        $result = $this->getEventManager()->trigger(MvcEvent::EVENT_DISPATCH, $e, function ($test) {
+        $result = $this->getEventManager()->triggerEventUntil(function ($test) {
             return ($test instanceof Response);
-        });
+        }, $e);
 
         if ($result->stopped()) {
             return $result->last();
--- a/src/Controller/ControllerManager.php
+++ b/src/Controller/ControllerManager.php
@@ -11,6 +11,7 @@ namespace Zend\Mvc\Controller;
 
 use Zend\EventManager\EventManagerAwareInterface;
 use Zend\EventManager\EventManagerInterface;
+use Zend\EventManager\SharedEventManagerInterface;
 use Zend\Mvc\Exception;
 use Zend\ServiceManager\AbstractPluginManager;
 use Zend\ServiceManager\ConfigInterface;
@@ -73,10 +74,10 @@ class ControllerManager extends AbstractPluginManager
             // is why the shared EM injection needs to happen; the conditional
             // will always pass.
             $events = $controller->getEventManager();
-            if (!$events instanceof EventManagerInterface) {
+            if (! $events instanceof EventManagerInterface
+                || ! $events->getSharedManager() instanceof SharedEventManagerInterface
+            ) {
                 $controller->setEventManager($parentLocator->get('EventManager'));
-            } else {
-                $events->setSharedManager($parentLocator->get('SharedEventManager'));
             }
         }
 
--- a/src/Controller/ControllerManager.php
+++ b/src/Controller/ControllerManager.php
@@ -10,7 +10,6 @@
 namespace Zend\Mvc\Controller;
 
 use Zend\EventManager\EventManagerAwareInterface;
-use Zend\EventManager\EventManagerInterface;
 use Zend\EventManager\SharedEventManagerInterface;
 use Zend\Mvc\Exception;
 use Zend\ServiceManager\AbstractPluginManager;
--- a/src/Controller/ControllerManager.php
+++ b/src/Controller/ControllerManager.php
@@ -74,9 +74,7 @@ class ControllerManager extends AbstractPluginManager
             // is why the shared EM injection needs to happen; the conditional
             // will always pass.
             $events = $controller->getEventManager();
-            if (! $events instanceof EventManagerInterface
-                || ! $events->getSharedManager() instanceof SharedEventManagerInterface
-            ) {
+            if (! $events || ! $events->getSharedManager() instanceof SharedEventManagerInterface) {
                 $controller->setEventManager($parentLocator->get('EventManager'));
             }
         }
--- a/src/Controller/Plugin/Forward.php
+++ b/src/Controller/Plugin/Forward.php
@@ -177,9 +177,9 @@ class Forward extends AbstractPlugin
             $results[$id] = [];
             foreach ($eventArray as $eventName => $classArray) {
                 $results[$id][$eventName] = [];
-                $events = $sharedEvents->getListeners($id, $eventName);
+                $events = $sharedEvents->getListeners([$id], $eventName);
                 foreach ($events as $currentEvent) {
-                    $currentCallback = $currentEvent->getCallback();
+                    $currentCallback = $currentEvent;
 
                     // If we have an array, grab the object
                     if (is_array($currentCallback)) {
@@ -193,7 +193,7 @@ class Forward extends AbstractPlugin
 
                     foreach ($classArray as $class) {
                         if ($currentCallback instanceof $class) {
-                            $sharedEvents->detach($id, $currentEvent);
+                            $sharedEvents->detach($currentEvent, $id);
                             $results[$id][$eventName][] = $currentEvent;
                         }
                     }
--- a/src/DispatchListener.php
+++ b/src/DispatchListener.php
@@ -99,8 +99,7 @@ class DispatchListener extends AbstractListenerAggregate
             $e->setControllerClass(get_class($controller));
             $e->setParam('exception', $ex);
 
-            $results = $events->triggerEvent($e);
-            $return  = $results->last();
+            $return = $events->triggerEvent($e)->last();
             if (! $return) {
                 $return = $e->getResult();
             }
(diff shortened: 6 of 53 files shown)


weierophinney:
Pinging @zendframework/community-review-team and @ezimuel — would love some review. :smile:

src/Application.php
```diff
@@ -103,15 +104,18 @@ class Application implements
      * @param mixed $configuration
      * @param ServiceManager $serviceManager
      */
-    public function __construct($configuration, ServiceManager $serviceManager)
-    {
+    public function __construct(
+        $configuration,
+        ServiceManager $serviceManager,
+        EventManagerInterface $events,
```
danizord, line 110: Should we provide defaults to avoid BC break?

Ocramius, line 110: Absolutely agree here.

src/Controller/AbstractController.php
```diff
@@ -109,13 +109,14 @@ public function dispatch(Request $request, Response $response = null)
         $this->response = $response;
 
         $e = $this->getEvent();
```
Ocramius, line 111: While at it, maybe rename it `$event`? ;-)

src/Controller/ControllerManager.php
```diff
@@ -73,10 +74,10 @@ public function injectControllerDependencies($controller, ServiceLocatorInterfac
             // is why the shared EM injection needs to happen; the conditional
             // will always pass.
             $events = $controller->getEventManager();
-            if (!$events instanceof EventManagerInterface) {
+            if (! $events instanceof EventManagerInterface
```
Ocramius: `if (! $events) {` is sufficient here, as per `EventManagerAwareInterface`

src/Controller/Plugin/Forward.php
```diff
@@ -193,7 +193,7 @@ protected function detachProblemListeners(SharedEvents $sharedEvents)
 
                     foreach ($classArray as $class) {
                         if ($currentCallback instanceof $class) {
-                            $sharedEvents->detach($id, $currentEvent);
+                            $sharedEvents->detach($currentEvent, $id);
```
Ocramius, line 196: This is a super-subtle bc break. Is it possible to revert this in the EVM instead?

src/Application.php
```diff
@@ -350,8 +359,12 @@ protected function completeRequest(MvcEvent $event)
...
         $event->setTarget($this);
-        $events->trigger(MvcEvent::EVENT_RENDER, $event);
-        $events->trigger(MvcEvent::EVENT_FINISH, $event);
+
+        $event->setName(MvcEvent::EVENT_RENDER);
+        $events->triggerEvent($event);
+
+        $event->setName(MvcEvent::EVENT_FINISH);
+        $events->triggerEvent($event);
         return $this;
```
Ocramius, line 369: Newline before this line (keeps consistent with the changes above)

src/Application.php
```diff
@@ -311,14 +318,16 @@ public function run()
...
         // Complete response
         $response = $result->last();
         if ($response instanceof ResponseInterface) {
+            $event->setName(MvcEvent::EVENT_FINISH);
             $event->setTarget($this);
             $event->setResponse($response);
-            $events->trigger(MvcEvent::EVENT_FINISH, $event);
+            $events->triggerEvent($event);
             $this->response = $response;
             return $this;
```
Ocramius: Newline before this line (keeps consistent with the changes above)

src/DispatchListener.php
```diff
@@ -92,12 +93,14 @@ public function onDispatch(MvcEvent $e)
...
-            $e->setError($application::ERROR_EXCEPTION)
-                  ->setController($controllerName)
-                  ->setControllerClass(get_class($controller))
-                  ->setParam('exception', $ex);
-            $results = $events->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $e);
-            $return = $results->last();
+            $e->setName(MvcEvent::EVENT_DISPATCH_ERROR);
+            $e->setError($application::ERROR_EXCEPTION);
+            $e->setController($controllerName);
+            $e->setControllerClass(get_class($controller));
```
Ocramius, line 99: Can this be part of the event class instead?

src/DispatchListener.php
```diff
@@ -92,12 +93,14 @@ public function onDispatch(MvcEvent $e)
...
+            $e->setParam('exception', $ex);
+
+            $results = $events->triggerEvent($e);
+            $return  = $results->last();
```
Ocramius: Can chain here

src/DispatchListener.php
```diff
@@ -211,12 +215,13 @@ protected function marshalBadControllerEvent(
         Application $application,
         \Exception $exception
     ) {
-        $event->setError($application::ERROR_EXCEPTION)
-              ->setController($controllerName)
-              ->setParam('exception', $exception);
+        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
+        $event->setError($application::ERROR_EXCEPTION);
```
Ocramius, line 218: Why a variable here btw?

src/DispatchListener.php
```diff
@@ -211,12 +215,13 @@ protected function marshalBadControllerEvent(
...
+        $event->setController($controllerName);
+        $event->setParam('exception', $exception);
```
Ocramius, line 220: We should probably make this a constant somewhere, so that we don't have users rely on a magic string (and we know when we're breaking BC)

src/DispatchListener.php
```diff
@@ -153,15 +156,16 @@ protected function marshalControllerNotFoundEvent(
         Application $application,
         \Exception $exception = null
     ) {
-        $event->setError($type)
-              ->setController($controllerName)
-              ->setControllerClass('invalid controller class or alias: ' . $controllerName);
+        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
+        $event->setError($type);
+        $event->setController($controllerName);
+        $event->setControllerClass('invalid controller class or alias: ' . $controllerName);
```
Ocramius, line 161: Eh, weird as heck. Strange that I never noticed this before :|

src/Controller/ControllerManager.php
Ocramius: Urgh, wait, the interface defines no getters... https://github.com/zendframework/zend-eventmanager/blob/a6c747fe59658fb62feb23613d7ea22146b6b611/src/EventManagerAwareInterface.php#L15-L24

src/Controller/ControllerManager.php
```diff
@@ -73,10 +74,10 @@ public function injectControllerDependencies($controller, ServiceLocatorInterfac
             // is why the shared EM injection needs to happen; the conditional
             // will always pass.
             $events = $controller->getEventManager();
```
Ocramius, line 75: Will need to check this against an interface (as per comment below), as the _aware_ interface defines no getter :-(

src/Service/ApplicationFactory.php
```diff
@@ -26,6 +26,12 @@ class ApplicationFactory implements FactoryInterface
      */
     public function createService(ServiceLocatorInterface $serviceLocator)
     {
-        return new Application($serviceLocator->get('Config'), $serviceLocator);
+        return new Application(
+            $serviceLocator->get('Config'),
+            $serviceLocator,
+            $serviceLocator->get('EventManager'),
```
Ocramius, line 32: As mentioned above, it may be a better idea to keep this as it was before

src/Service/ServiceManagerConfig.php
```diff
@@ -94,15 +95,20 @@ public function __construct(array $configuration = [])
     {
         $this->initializers = [
             'EventManagerAwareInitializer' => function ($instance, ServiceLocatorInterface $serviceLocator) {
-                if ($instance instanceof EventManagerAwareInterface) {
-                    $eventManager = $instance->getEventManager();
+                if (! $instance instanceof EventManagerAwareInterface) {
+                    return;
+                }
+
+                $eventManager = $instance->getEventManager();
```
Ocramius, line 102: Same as above. The `aware` interface defines no getter.

src/Controller/AbstractController.php
gianarb, line 111: :+1:

src/View/Http/DefaultRenderingStrategy.php
```diff
@@ -108,9 +108,11 @@ public function render(MvcEvent $e)
 
             $application = $e->getApplication();
             $events      = $application->getEventManager();
-            $e->setError(Application::ERROR_EXCEPTION)
-              ->setParam('exception', $ex);
-            $events->trigger(MvcEvent::EVENT_RENDER_ERROR, $e);
+
+            $e->setError(Application::ERROR_EXCEPTION);
+            $e->setParam('exception', $ex);
```
Ocramius, line 113: Same as above: rely on a constant instead?

test/ApplicationTest.php
```diff
@@ -69,15 +74,13 @@ public function setUp()
 
     public function getConfigListener()
     {
-        $manager   = $this->serviceManager->get('ModuleManager');
-        $listeners = $manager->getEventManager()->getListeners('loadModule');
-        foreach ($listeners as $listener) {
-            $callback = $listener->getCallback();
-            if (!is_array($callback)) {
+        $manager = $this->serviceManager->get('ModuleManager');
+        foreach ($this->getListenersForEvent('loadModule', $manager->getEventManager()) as $listener) {
```
Ocramius: Do we have a constant for this event name?

test/EventManagerIntrospectionTrait.php
```diff
@@ -0,0 +1,56 @@
...
+
+use ReflectionProperty;
+use Zend\EventManager\EventManager;
+
+/**
+ * Offer methods for introspecting event manager events and listeners.
+ */
+trait EventManagerIntrospectionTrait
+{
+    public function getEventsFromEventManager(EventManager $events)
```
Ocramius: Why can't this be a static method somewhere (instead of a trait)? Also: `private`

test/EventManagerIntrospectionTrait.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $r = new ReflectionProperty($events, 'events');
+        $r->setAccessible(true);
+        $listeners = $r->getValue($events);
+        return array_keys($listeners);
+    }
+
+    public function getListenersForEvent($event, EventManager $events, $withPriority = false)
```
Ocramius: Same as above

test/EventManagerIntrospectionTrait.php
```diff
@@ -0,0 +1,56 @@
...
+        $listeners = $r->getValue($events);
+
+        if (! isset($listeners[$event])) {
+            return $this->traverseListeners([]);
+        }
+
+        return $this->traverseListeners($listeners[$event], $withPriority);
+    }
+
+    public function traverseListeners(array $queue, $withPriority = false)
```
Ocramius, line 121: Same as above

Ocramius, line 121: Method purpose is unclear

test/View/Console/DefaultRenderingStrategyTest.php
```diff
@@ -18,10 +18,13 @@
...
 use Zend\Stdlib\Response;
 use Zend\View\Model;
+use ZendTest\Mvc\EventManagerIntrospectionTrait;
 
 class DefaultRenderingStrategyTest extends TestCase
 {
-    /** @var DefaultRenderingStrategy */
+    use EventManagerIntrospectionTrait;
+
+    /* @var DefaultRenderingStrategy */
```
Ocramius: Should use `/**`, not `/*`

test/View/RouteNotFoundStrategyTest.php
```diff
@@ -288,35 +291,33 @@ public function testNotFoundTemplateIsMutable()
...
+            $expectedListener = [$this->strategy, 'prepareNotFoundViewModel'];
             $found            = false;
-            foreach ($listeners as $listener) {
-                $callback = $listener->getCallback();
-                if ($callback === $expectedCallback) {
-                    if ($listener->getMetadatum('priority') == $expectedPriority) {
-                        $found = true;
-                        break;
-                    }
+            foreach ($listeners as $priority => $listener) {
```
Ocramius: This piece of logic should probably be part of the static/trait above

test/View/RouteNotFoundStrategyTest.php
```diff
@@ -16,9 +16,12 @@
 use Zend\Mvc\MvcEvent;
 use Zend\Mvc\View\Http\RouteNotFoundStrategy;
 use Zend\View\Model\ViewModel;
+use ZendTest\Mvc\EventManagerIntrospectionTrait;
 
 class RouteNotFoundStrategyTest extends TestCase
 {
+    use EventManagerIntrospectionTrait;
```
Ocramius, line 23: After reading through the tests, I'm more convinced that the trait should just be a static class

src/Controller/Plugin/Forward.php
weierophinney, line 196: No. In the new version, the second argument is optional, the first required.

src/DispatchListener.php
weierophinney, line 99: That's out of scope for this particular refactor.

src/Application.php
gianarb, line 110: This is a BC break but we are working to mvc 3.0 if it is a GOOD BC break this is a good time to merge it

src/DispatchListener.php
weierophinney, line 218: Can't remember why, TBH; that's old code.

src/Application.php
Ocramius, line 110: @gianarb it is kind of a big one, tbh, and it can be avoided

src/Controller/ControllerManager.php
weierophinney: Yes it does! It inherits from EventsCapableInterface, which defines the
getter.

src/DispatchListener.php
Ocramius, line 99: Alright, will raise it as an issue after we get to the end with this PR

src/Controller/ControllerManager.php
weierophinney, line 75: Please see my previous comment...

src/Controller/Plugin/Forward.php
Ocramius, line 196: I see. Still very very very annoying BC break

src/View/Http/DefaultRenderingStrategy.php
weierophinney, line 113: Out of scope for this refactor. I am literally trying to do the bare
minimum to get the component working with the new EM version. Open a
separate PR for such changes, if you feel they're important.

Ocramius, line 113: Will spawn an issue

src/Service/ApplicationFactory.php
weierophinney, line 32: Add noted in my issue summary, I needed to make this change due to a race
condition with EM injection.

Since most people never directly instantiate, this will affect very few
people. I'll be sure to call it out in the migration guide.

test/EventManagerIntrospectionTrait.php
weierophinney: I know you hate traits, but I found it an expedient way to accomplish the
task, and this is precisely the dirt of thing they were designed for.

I'll update to private, though I don't think the visibility is important
within a unit test.

test/EventManagerIntrospectionTrait.php
weierophinney, line 121: It's a generator. Suggest a better name?

test/View/Console/DefaultRenderingStrategyTest.php
weierophinney: I saw somebody content recently that these hints shouldn't be formal
docblocks. Can you link me to something showing otherwise?

test/View/RouteNotFoundStrategyTest.php
weierophinney, line 23: Why? What rationale can you provide?

test/View/Console/DefaultRenderingStrategyTest.php
Ocramius: @weierophinney `/**` => docblock (stored in reflection data as well), while `/*` is only for IDE hints.

Here's the practical example: https://3v4l.org/O3CgU

And copying it for reference:

``` php
<?php

class Foo
{
    /* stuff */
    private $bar;
    /** more stuff */
    private $baz;
}

var_dump((new ReflectionProperty(Foo::class, 'bar'))->getDocComment());
var_dump((new ReflectionProperty(Foo::class, 'baz'))->getDocComment());
```

src/Controller/ControllerManager.php
Ocramius, line 75: :+1:

src/Controller/ControllerManager.php
Ocramius: :+1:

src/DispatchListener.php
Ocramius, line 218: Alright, will flag up

src/Service/ApplicationFactory.php
Ocramius, line 32: Works for me.

test/EventManagerIntrospectionTrait.php
Ocramius: > I know you hate traits

I don't hate traits, I just see a static method as a better solution here. After all, these are helper functions

test/EventManagerIntrospectionTrait.php
Ocramius, line 121: What does the traversal do? Does it simply allow traversing the listeners by their priority, emulating what the event manager does?

Also, can the argument simply be the event manager (so that fetching the listeners becomes a detail)?

test/View/RouteNotFoundStrategyTest.php
Ocramius, line 23: As mentioned above, you are copying the helper methods into every test case, instead of just having them available in the test suite.

Since they are pure helper methods, you can just have them as static methods in a separate class.

test/EventManagerIntrospectionTrait.php
weierophinney, line 121: It's the listeners for a specific event. In terms of the priority flag:
- false should be used if you want to potentially cast to an array. An
  interesting thing with generators and iterators is that you can have keys
  that are identical - but iterator_to_array will collapse those. As such,
  this option does not provide keys.
- true should be used if you'll actually iterate, allowing you to get keys
  and listeners.

I could collapse this method into the getListenersForEvent() method,
though; will do so shortly.

test/View/RouteNotFoundStrategyTest.php
weierophinney, line 23: Doing so makes the verbiage far more verbose, though:

``` php
$this->getListenersForEvent(...);

// vs
EventManagerIntrospection::getListenersForEvent(...)
```

Let's leaves as-is, with the improvements and additions you've suggested.
I'll get those pushed shortly.

Ocramius, line 23: Alright

test/EventManagerIntrospectionTrait.php
Ocramius, line 121: I'm fine with a descriptive docblock as well, btw

src/Application.php
weierophinney, line 110: Putting a note here as threads got broken up by me answering via email.

This change was for a reason, which I stated in the PR summary: I ended up with a circular reference issue with the ApplicationFactory + EventManagerAware initializer that resulted in invalid EventManager state. The result was that the EM instance injected no longer had the default listeners injected, which broke the entire thing.

As I noted in another comment elsewhere: hardly anybody directly instantiates the `Application` instance. The skeleton application pulls it from the service manager. As such, we can call this change out in the migration documentation for the very small percentage of users who are manually instantiating.

src/Controller/AbstractController.php
weierophinney, line 111: Out of scope for this PR; I tried very hard only to update the code necessary to get the EM refactor working with it. If you feel strongly, open new PRs.

src/Controller/Plugin/Forward.php
weierophinney, line 196: It's a rarely used feature, TBH; there are very few use cases currently for detaching listeners, much less shared listeners, and the `Forward` plugin here is literally the only place I've seen it in the ZF2 or Apigility code bases.

Though with React, I suspect it will becomes more common. Those users will adopt v3 anyways, as it's faster. :smile:

src/Application.php
weierophinney, line 369: Noted, and staged locally to push with other changes.

src/DispatchListener.php
weierophinney: Chain which? I purposely didn't chain the setters for the event, as the `EventInterface` no longer defines fluent setters (though `MvcEvent` still does). Or did you mean this?

``` php
$return = $events->triggerEvent($e)->last();
```

src/Service/ServiceManagerConfig.php
weierophinney, line 102: For those not following the whole thread, I answered this: `EventManagerAwareInterface` extends `EventManagerCapableInterface`, which defines the getter.

test/ApplicationTest.php
weierophinney: Yep: `ModuleEvent::EVENT_LOAD_MODULE`. I'll update the code.

test/View/Console/DefaultRenderingStrategyTest.php
weierophinney: I missed that it was an **object** property. My bad!

test/View/RouteNotFoundStrategyTest.php
weierophinney: Working on that now. It's actually an argument for a trait, as then I can do assertions within a method. I'm creating `assertListenerAtPriority(callable $expectedListener, $expectedPriority, $event, EventManager $events, $message = '')` right now; greatly reducing a ton of this boilerplate, and making the test cases easier to understand.

weierophinney:
Feedback incorporated, @Ocramius :smile:

weierophinney:
Bah, CS checks! Fixed and pushed.

Ocramius:
Thanks @weierophinney!

composer.json
```diff
@@ -14,11 +14,12 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-eventmanager": "~2.5",
+        "zendframework/zend-eventmanager": "dev-develop as 2.7.0",
```
Ocramius, line 17: Note: should probably use `zend-eventmanager dev-develop as 3.0.0`

weierophinney, line 17: Can't yet; not all components are updated to zend-eventmanager v3 yet, which means that there will be an unresolveable conflict. (I know; I tried. :smile:)

weierophinney:
Added a migration guide covering the BC breaks introduced, and updated the branch-alias for the develop branch to 3.0-dev.

weierophinney:
Okay, merging. If anybody finds anything else wrong, please open an issue or PR and ping me. :smile:

PR #32, opened 2015-10-01
Add a middleware dispatch

ezimuel:
**NOTE:** This PR is a WIP, please don't merge it.

This PR adds the middleware support to MVC. We want to be able to dispatch a "middleware" parameter as callable or service, using the signature:

``` php
(
  Psr\Http\Message\ServerRequestInterface $request,
  Psr\Http\Message\ResponseInterface $response
)
```

To do:
- [x] Fix and complete the unit test

### DIFF
--- a/src/DispatchListener.php
+++ b/src/DispatchListener.php
@@ -14,8 +14,6 @@ use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface;
 use Zend\Mvc\Exception\InvalidControllerException;
 use Zend\Stdlib\ArrayUtils;
-use Zend\Psr7Bridge\Psr7ServerRequest as Psr7Request;
-use Zend\Psr7Bridge\Psr7Response;
 
 /**
  * Default dispatch listener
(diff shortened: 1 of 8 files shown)


src/DispatchListener.php
```diff
@@ -61,11 +63,30 @@ public function attach(EventManagerInterface $events)
...
+            }
+            if (!is_callable($middleware)) {
+                $middleware = is_string($middleware) ? $middleware : get_class($middleware);
+                $return = $this->marshalControllerNotFoundEvent($application::ERROR_MIDDLEWARE_CANNOT_DISPATCH, $middleware, $e, $application);
+                return $this->complete($return, $e);
+            }
+            $return = $middleware(Psr7Request::fromZend($request), Psr7Response::fromZend($response));
+            return $this->complete(Psr7Response::toZend($return), $e);
+        }
+
```
weierophinney: I would argue we should separate this logic into a _different_ listener, which we would then add as a default listener in `Zend\Mvc\Application`; it would execute at higher priority than this one (or same priority, but register prior to this one). If no middleware route match parameter is found, it would simply return, which would cause the original dispatch listener to take over.

weierophinney: I'd write the end logic as this:

``` php
if (! $return instanceof \Psr\Http\Message\ResponseInterface) {
    $e->setResult($return);
    return $return;
}
$response = Psr7Response::toZend($return);
$e->setResult($response);
return $response;
```

which might fix the testing issue you're seeing.

ezimuel:
@weierophinney I created a MiddlewareListener with the logic suggested but I'm having issue to trigger this events. The unit tests are failing, the DispatchListener is executed before the MiddlewareListener even if I used a priority of 1000 [here](https://github.com/ezimuel/zend-mvc/commit/47ac19ad8d09936fcc729b82ae1c01bb724e83a7#diff-7dbb59d47c1c639fd43c2fe2cf961672R30).

src/MiddlewareListener.php
```diff
@@ -0,0 +1,116 @@
...
+{
+    /**
+     * Attach listeners to an event manager
+     *
+     * @param  EventManagerInterface $events
+     * @return void
+     */
+    public function attach(EventManagerInterface $events)
+    {
+        $this->listeners[] = $events->attach(MvcEvent::EVENT_DISPATCH, [$this, 'onDispatch'], 1000);
```
weierophinney: We'll likely take this down to 1 once we solve the issue.

ezimuel:
@weierophinney I fixed the unit test issue. The test runs with two errors on `ZendTest\Mvc\Service\TranslatorServiceFactoryTest`. I guess that is caused for the refactor of zend-eventmanager.

weierophinney:
You're using dev-develop of the ModuleManager, which is currently setup to use dev-develop of the servicemanager — which is the source of that issue. You can ignore it for now.

src/MiddlewareListener.php
```diff
@@ -0,0 +1,116 @@
...
+              ->setControllerClass(get_class($middleware))
+              ->setParam('exception', $ex);
+            $results = $events->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $e);
+            $return = $results->last();
+            if (! $return) {
+                $return = $e->getResult();
+            }
+        }
+
+        if (! $return instanceof \Psr\Http\Message\ResponseInterface) {
```
samsonasik, line 75: move to use statement, I created PR to your branch for it https://github.com/ezimuel/zend-mvc/pull/1

PR #33, opened 2015-10-04
sorted use statements

EmanueleMinotto:
My first PR: no real changes but a more tidy source code. I used https://github.com/mmoreram/php-formatter

If no automation can be applied than I can send a PR for all the other repositories, let me know.
Thanks


Maks3w:
Could you create a PHPCS Fixer for automatize this?

Maks3w:
I'll close this because PHP-CS-Fixer had the option and we never opt for enable that check.

PR #34, opened 2015-10-06
Fix @return comment of getRequest()  for correct auto completion

marinovdf:


Maks3w:
`HttpRequest` is an implementation of `Zend\Stdlib\RequestInterface`

marinovdf:
Yes, i understand. 
PhpStorm says "Method `getPost` not found in class `\Zend\Stdlib\RequestInterface`", if I write `$this->getRequest()->getPost()` in my controller. Maybe it's bug of PhpStorm.
Sorry for my English.

manuakasam:
There is no bug.

The implementation by ZF2 simply offers much more than the interface is
requiring. That is not an error or bug at all.

What you can do is this:

```
/** @var \Zend\Http\PhpEnvironment\Request $request */
$request = $this->getRequest();
```

And with this you'll then have full IDE support of all the functions :)

marinovdf notifications@github.com schrieb am Di., 6. Okt. 2015 um
12:44 Uhr:

Maks3w:
IIRC its possible add the magic `@method` annotation at class level. However for to be purist the best option is use the `instanceof` operator with the return of `getRequest` and avoid accidental PHP Errors.

PR #36, opened 2015-10-08
Refactored to work with zend-servicemanager v3

weierophinney:
This patch updates zend-mvc to use the upcoming zend-servicemanager v3.
- All factories implementing `FactoryInterface` and/or `AbstractFactoryInterface` were updated to the new interface signatures.
- All plugin managers were updated to the new `AbstractPluginManager` changes.  Whenever possible, they were rewritten to remove invokables and instead define factories and/or aliases.
- Added the `RouteInvokableFactory`, which acts both as a specialized invokable factory for route implementations (as they use a `::factory()` method for initialization vs a constructor), and an abstract factory for handling FQCN route names. The `RoutePluginManager` composes this abstract factory by default.
- `Application::init` was updated to pull the fully configured service manager from the `ServiceListener` after loading modules.
- The `DispatchListener` now receives its controller manager instance during instantiation.
- Separated router factory into 3:
  - `ConsoleRouterFactory` will create a console router
  - `HttpRouterFactory` will create an HTTP router
  - `RouterFactory` delegates to the above two, based on status of `Console::isConsole()`
  
  The above makes it possible to specify the specific router type you want to use via configuration, making testing possible.
- Refactored the ViewManager implementations. Previously, the view managers were adding services and aliases to the service manager duing their bootstrap listeners; with the SM now immutable, that does not make sense, and, in fact, leads to errors.
  
  In all cases where object creation was occuring, the code was moved to new factories, which the view manager implementations can now pull specifically.  Aliases for these were added where appropriate. However, I stopped short of adding factories for general services such as `DefaultRenderingStrategy` as the implementations differ enough that having such a factory mutate based on the selected view manager would require type-hinting to consume the services regardless.
- New integration tests for `Application` were written to ensure the SM is in correct state for `bootstrap()` and `run()`, and that the service listener <-> service manager dance is correctly managed.


weierophinney:
Decrease in test coverage is principally from disabling a number of tests due to the components they integrate with not yet being updated to zend-servicemanager v3 (notably: zend-i18n and zend-form). In most cases, I've _added_ tests.

weierophinney:
Ping @zendframework/community-review-team — finally have the zend-servicemanager integration done!

src/Router/RouteInvokableFactory.php
```diff
@@ -0,0 +1,83 @@
...
+     * @param string $routeName
+     * @return bool
+     */
+    public function canCreateServiceWithName(ContainerInterface $container, $routeName)
+    {
+        if (! class_exists($routeName)) {
+            return false;
+        }
+
+        if (! is_subclass_of($routeName, RouteInterface::class)) {
```
samsonasik, line 39: how about using `$routeName instanceof RouteInterface` ?

Maks3w, line 39: $routeName is string

weierophinney, line 39: We cannot use instanceof as $routeName is the string class name, not an
object instance.

src/Service/ServiceListenerFactory.php
```diff
@@ -81,24 +85,35 @@ class ServiceListenerFactory implements FactoryInterface
...
             'Zend\Mvc\Controller\PluginManager'          => 'ControllerPluginManager',
             'Zend\Mvc\View\Http\InjectTemplateListener'  => 'InjectTemplateListener',
+            'Zend\View\Renderer\RendererInterface'       => 'Zend\View\Renderer\PhpRenderer',
             'Zend\View\Resolver\TemplateMapResolver'     => 'ViewTemplateMapResolver',
             'Zend\View\Resolver\TemplatePathStack'       => 'ViewTemplatePathStack',
             'Zend\View\Resolver\AggregateResolver'       => 'ViewResolver',
             'Zend\View\Resolver\ResolverInterface'       => 'ViewResolver',
-            'ControllerManager'                          => 'ControllerLoader',
         ],
+        /*
```
Maks3w, line 112: Can be this removed?

weierophinney, line 112: I'm still trying to decide what to do with it; forms will likely be optional, but we don't have a documented way to register the abstract factory yet. Once I have a plan, yes, we'll likely remove it.

PR #37, opened 2015-10-09
update for zend\hydrator 2.0

prolic:


Maks3w:
There is no reason for raise the minimum required version

PR #40, opened 2015-10-23
update Redirect::toRoute() docblock as not only generate url

samsonasik:
but redirect based on url that generated based on given route.

### DIFF
--- a/src/Controller/Plugin/Redirect.php
+++ b/src/Controller/Plugin/Redirect.php
@@ -23,7 +23,7 @@ class Redirect extends AbstractPlugin
     protected $response;
 
     /**
-     * Generates a URL based on a route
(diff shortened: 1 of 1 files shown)


Maks3w:
I agree the original comment wasn't correct but the proposed is not enough precise. 

The function don't perform any redirection, it's generate a redirect response. 

This difference is important because with the proposed description developer may understand the request end and no more code will be executed (lije zf1 goToRouteAndExit) but it still require the developer to do something more with the generated response

samsonasik:
So, it can be changed to: "Generate redirect response based on given route" ? what about `toUrl()`, need update too?

Maks3w:
Yes to both questions

samsonasik:
done, I've updated both

Issue #42, opened 2015-10-30
Imposible to use raw data in AbstractRestfulController in some cases


svycka:
here https://github.com/zendframework/zend-mvc/blob/8a464ff52049a6ad4990347eb30f69fb79f14a47/src/Controller/AbstractRestfulController.php#L586 is unreachable code

```
parse_str($content, $parsedParams); // returned $parsedParams keys not indexes
```

there are only one way to get raw response if $content is something like `0=something` otherwise if `$content = 'foo'` or similar parse_str() will return `['foo' => '']`

svycka:
for example if you use:

```
public function update($name, $data) {
    // here $data is always array but sometimes I need just a string or xml or something.. I want raw data
}
```

svycka:
ping @weierophinney or someone any response?

weierophinney:
Fixed with #43

PR #43, opened 2015-11-12
BUG FIX: AbstractRestfulController does not allow xml, yml(or any string) as body

svycka:
more info https://github.com/zendframework/zend-mvc/issues/42


svycka:
This is not perfect solution so I am open to suggestions how to fix this if you don't like this fix.

svycka:
ping @weierophinney @Maks3w

PR #47, opened 2015-12-01
change withConfig() call with configure()

samsonasik:
in favor of https://github.com/zendframework/zend-servicemanager/pull/47


weierophinney:
This is not forwards/backwards compatible with both v2 and v3. The appropriate solution is to use a `Zend\ServiceManager\Config` instance, and have it call `configure()`, passing the plugin manager.

samsonasik:
it cherry-picked in #48 by @mtymek could you take care of that?

PR #55, opened 2016-01-04
[application] Safe retrieve of request service

Maks3w:
Request service is made with data from external input (like browser data or command line arguments)

If the input is malformed (accidentally or intentionally) the Request service cannot be created and a non catched exception is throw killing the application.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -12,6 +12,7 @@ namespace Zend\Mvc;
 use Zend\EventManager\EventManagerAwareInterface;
 use Zend\EventManager\EventManagerInterface;
 use Zend\ServiceManager\ServiceManager;
+use Zend\Stdlib\RequestInterface;
 use Zend\Stdlib\ResponseInterface;
 
 /**
@@ -83,19 +84,19 @@ class Application implements
     protected $events;
 
     /**
-     * @var \Zend\Stdlib\RequestInterface
+     * @var null|RequestInterface
      */
-    protected $request;
+    private $request;
 
     /**
-     * @var ResponseInterface
+     * @var null|ResponseInterface
      */
-    protected $response;
+    private $response;
 
     /**
      * @var ServiceManager
      */
-    protected $serviceManager = null;
+    private $serviceManager;
 
     /**
      * Constructor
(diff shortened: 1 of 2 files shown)


weierophinney:
Also marking as BC break, as it makes the request/response properties private (which could break any implementations that extend the class). That said, reasonable break, and will resolve some issues.

weierophinney:
@Maks3w can you please rebase?

src/Application.php
```diff
@@ -149,19 +149,30 @@ public function bootstrap(array $listeners = [])
...
+            $event->setRequest(new Request());
+            $event->setError(self::ERROR_EXCEPTION);
+            $event->setParam('exception', $e);
+        }
+
         // Trigger bootstrap events
-        $events->trigger(MvcEvent::EVENT_BOOTSTRAP, $event);
+        if ($event->isError()) {
+            $events->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $event);
+            $this->completeRequest($event);
```
weierophinney, line 166: I think this would require a new event; since we're not done with bootstrap, it's entirely possible that the application is not in a fully usable state, and triggering a dispatch.error event could lead to cascading errors.

weierophinney:
I just spent a bit of time rebasing this and attempting to get tests to pass, but feel at this time, the approach is incomplete.

The issue is that this happens during bootstrap, which leads to compounding problems:
- Each of `DISPATCH_ERROR`, `RENDER`, and `FINISH` are triggered, but because bootstrap has not yet complete, these listeners may not be completely setup / configured, leading to unexpected results.
- On top of that, even on exception, the method returns normally, but `run()` has no logic present to understand that the application _should not be run_ due to inability to complete bootstrapping. This will lead to additional errors, particularly as the request instance is now in a known incomplete/invalid state for most operations.

I see two possibilities to make this work more correctly.
- Trigger an alternate event that will deliver static contents on failure to create a request instance, and mark the event as "complete" or similar. `run()` would then return early when such a condition is met.
- Modify bootstrapping to do a try / catch block around request creation, and return a static page for exceptions, bypassing the application entirely. This logic would need to occur _before_ attempting to pull the `Application` instance, or we would need to modify the `Application` instance to lazy-fetch the request on-demand (as the first commit to this patch does).

For now, I'm closing this; if somebody wants to make this happen, please see my above notes.

PR #57, opened 2016-01-18
[ZF3] remove @deprecated classes and methods

samsonasik:


weierophinney:
Please re-issue this _after_ the items in https://github.com/zendframework/maintainers/wiki/zend-mvc-v3-refactor:-reduce-components are complete; until then, there's a lot of duplication, and some of this is being treated differently based on removal of components as dependencies anyways.

Thanks!

PR #63, opened 2016-01-28
Add composer.lock into the source and update Travis build matrix

stefanotorresi:
As discussed in zendframework/zf2#7660 this PR does the following:
- removes `composer.lock` from `.gitignore`
- compile the lock file against dependencies available at the time of this PR
- updates `.travis.yml` to build against three dependencies status: `lowest`, `locked` and `latest` (expressed by the `DEPS` environment variable)

Since I'm at it, I'm also proposing the following changes, but we've not discussed these yet so I wait for CR to validate them:
- add common scripts in `composer.json`, like it has been done in  [zend-expressive](https://github.com/zendframework/zend-expressive/blob/325fbb36ed4c0b04267aa25217ff999b48c48e98/composer.json#L59-L68)
- move php 7 up in the travis matrix, so it won't be allowed to fail anymore, and use it for checking cs and testing against latest deps. Doing so, a considerable speed boost can be appreciated in the build (which now runs in half the time and retains the same number of jobs). Besides, we'll have to bump it sooner or later, won't we? Nevertheless, if you think it's premature, I'll move the `DEPS=latest; CHECK_CS=true` env in a new 5.6 job. This shouldn't cause the build to run any slower than the current `master` one though, but of course it will be somewhat slower (~100 secs) than the one relative to this PR.

The new `DEPS=lowest` build environment already shows the following dependency issues:
- PHPUnit versions `<4.0.2|>=4.1` installed from tarball (i.e. not with `--prefer-source`) AND with `--prefer-lowest` will come with `sebastian/version:1.0.0`, which has a bug causing phpunit to report the wrong version number. This makes ZF phpunit bootstrap scripts fail across every component, because they validate this version. This is definitely an upstream issue, but it can be easily fixed by explicitly requiring `sebastian/version:^1.0.3`.
- PHPUnit 4.0.\* causes fatals in some tests when mocking interfaces. Using 4.1 fixes this issue. Current `composer.json` doesn't reflect this requirement.
- `master` branch of `zend-mvc` is not compatible with `zend-modulemanager` version 2.6, but actually requires 2.6.1 to pass the tests. Current `composer.json` doesn't reflect this requirement.

If the current changes are approved, I'll move on to replicate the same changes for the develop branch of this repo and all the other framework components.


stefanotorresi:
ping @weierophinney @Ocramius

weierophinney:
@stefanotorresi — Can you rebase and re-submit this against the current develop branch please?

Also, a few notes:
- develop is already requiring PHP 7 builds. :smile:
- problems with the zend-modulemanager versions are also corrected on the develop branch
- the issues you describe with PHPUnit when using `--prefer-lowest` are why we've been using explicit versions when testing (see the travis config on the develop branch). If you can resolve those with changes in the dev tool versions, that'd be fantastic, but if not, then the approach we're using is well-vetted at this point. (It's also why I'm not sure if a `--locked` will work)

PR #79, opened 2016-02-29
Composer lock

stefanotorresi:
resubmit of #63 targeting `develop` branch.

@weierophinney it worked like a charm :) the `DEPS=lowest` env does exactly what was being done with the package env vars, only now it looks simpler because it comes down to how the deps are defined in `composer.json`, rather than overriding each one of them in `.travis.yml`.
The PHPUnit requirement needed to be bumped up to `^4.5` because one of the tests was using a Prophecy method, which has been introduced with that version.

Let me know if you're satisfied and I'll proceed applying the same changes to all the components.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,10 +89,10 @@ install:
 script:
   - if [[ $TEST_COVERAGE == 'true' ]]; then composer test-coverage ; else composer test ; fi
   - if [[ $CHECK_CS == 'true' ]]; then composer cs-check ; fi
+  - if [[ $DEPLOY_DOCS == "true" && "$TRAVIS_TEST_RESULT" == "0" ]]; then travis_retry curl -sSL https://raw.githubusercontent.com/zendframework/zf-mkdoc-theme/master/theme-installer.sh | bash ; fi
 
 after_script:
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer upload-coverage ; fi
 
 after_success:
-  - if [[ $DEPLOY_DOCS == "true" ]]; then travis_retry curl -sSL https://raw.githubusercontent.com/zendframework/zf-mkdoc-theme/master/theme-installer.sh | bash ; fi
   - if [[ $DEPLOY_DOCS == "true" ]]; then ./zf-mkdoc-theme/deploy.sh ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ before_install:
 install:
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
-  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS satooshi/php-coveralls ; fi
+  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --no-update --dev $COMPOSER_ARGS satooshi/php-coveralls ; fi
   - travis_retry composer install $COMPOSER_ARGS
   - composer show --installed
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,6 @@ cache:
 
 env:
   global:
-    - EVENT_MANAGER_VERSION="^3.0"
-    - HYDRATOR_VERSION="^2.1"
-    - SERVICE_MANAGER_VERSION="^3.0.3"
-    - STDLIB_VERSION="^3.0"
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -11,68 +11,72 @@ cache:
...
-  - composer require --no-update "zendframework/zend-eventmanager:$EVENT_MANAGER_VERSION" "zendframework/zend-servicemanager:$SERVICE_MANAGER_VERSION" "zendframework/zend-hydrator:$HYDRATOR_VERSION" "zendframework/zend-stdlib:$STDLIB_VERSION"
+  - travis_retry composer self-update
+  - if [[ $TRAVIS_PHP_VERSION != "hhvm" && $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini ; fi
 
 install:
-  - travis_retry composer install --no-interaction --ignore-platform-reqs
+  - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
+  - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
+  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS satooshi/php-coveralls ; fi
+  - travis_retry composer install $COMPOSER_ARGS
```
weierophinney, line 86: I'm a bit confused by the fact that we have this _after_ the update commands. While I can see that it runs fine on Travis, it's confusing.

What's the rationale for doing an install after an update? Shouldn't it be the other way around?

stefanotorresi, line 86: That is intended as a "catch-all" scenario: the updates are only executed in special cases, and executing an install right after an update is basically a no-op. Since every line of the `install` block represents a different build environment, I preferred to not enclose this one in a conditional block to make it stand out as the lowest common denominator.

svycka, line 86: @stefanotorresi then maybe add `--lock` to `composer update` to just update lock file if install is called in all scenarious?

.travis.yml
```diff
@@ -11,68 +11,72 @@ cache:
...
-  - if [[ $EXECUTE_TEST_COVERALLS == 'true' ]]; then composer require --dev --no-update satooshi/php-coveralls ; fi
-  - composer require --no-update "zendframework/zend-eventmanager:$EVENT_MANAGER_VERSION" "zendframework/zend-servicemanager:$SERVICE_MANAGER_VERSION" "zendframework/zend-hydrator:$HYDRATOR_VERSION" "zendframework/zend-stdlib:$STDLIB_VERSION"
+  - travis_retry composer self-update
+  - if [[ $TRAVIS_PHP_VERSION != "hhvm" && $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini ; fi
 
 install:
-  - travis_retry composer install --no-interaction --ignore-platform-reqs
+  - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
+  - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
+  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS satooshi/php-coveralls ; fi
```
svycka: why you removed `--no-update` here?

.travis.yml
stefanotorresi, line 86: @svycka good call! I forgot about that option!

.travis.yml
stefanotorresi: definitely an oversight, thanks :)

.travis.yml
stefanotorresi, line 86: @svycka apparently `--lock` doesn't play well with `--prefer-lowest`, so I'm gonna have to revert that.

weierophinney:
@stefanotorresi It looked like you were still working on this when I was prepping the 2.7 release, so I didn't get it in.

That said, I'm trying the approach with zend-test right now, and it's working brilliantly, so I'm definitely in favor! My only concern is the proliferation of test jobs; since Travis only allows 5 in parallel at any given time, this means the tests require at least two passes by Travis before completion. However, it ensures that code that purports to support multiple versions truly supports them, so it's a reasonable tradeoff in my opinion.

stefanotorresi:
@weierophinney rebased against latest develop to include the docs deploy steps. I've made some slight changes to those too, so please make sure they still work. :)
Also see zendframework/zf-mkdoc-theme#6 which is related.

stefanotorresi:
on a side note, if we _really_ wanted to reduce the jobs, maybe testing with lowest and latest deps on both 5.5 and 5.6 is not that useful, so we could get rid of 2 jobs, but I'm not sure it would do any good, because there would still be more than 5 jobs.

Issue #87, opened 2016-03-02
Deprecation warning


metanav:
I am getting following warning after upgrading zend-mvc:
_ServiceLocatorAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceLocatorAwareInitializer. Please update your class My\Controller\MyController to remove the implementation, and start injecting your dependencies via factory instead._

I have read the CHANGELOG and I understand why this warning is coming. But none of the controller classes is implementing ServiceLocatorAwareInterface in my application. I think due to extending AbstractController class which has setServiceLocator() method I am getting this warning. I know if I need I can inject it through factory. I would prefer to see this warning only if any classes is implementing ServiceLocatorAwareInterface, not for merely defining a method setServiceLocator() which can be there even if we inject it using setters in factory instead of constructor injection.

froschdesign:
Here is the problem: https://github.com/zendframework/zend-mvc/blob/master/src/Controller/ControllerManager.php#L223

metanav:
That's my concern. Now there is no implementation of ServiceLocatorAwareInterface so why this warning?

froschdesign:
@metanav 
I updated my previous comment. (Sorry, I was too slow!)

froschdesign:
I think, the warning is good, but message is wrong.

metanav:
I agree with you regarding the message. But could you please explain the logic behind triggering this warning when  you have a method  setServiceLocator in your controller?   Suppose I have controller and factory like this:
(Note: Neither I am extending AbstractController class or implementing ServiceLocatorAwareInterface)

```
class MyControllerFactory
 {
    public function __invoke($serviceLocator)
    {
        $myController = new MyController();
        $myController->setServiceLocator($serviceLocator);
    }
}

class MyController
{
    public function __constructor() {}
    public function setServiceLocator($serviceLocator) 
    {
         $this->serviceLocator = $serviceLocator;
    }
}
```

froschdesign:
>  But could you please explain the logic behind triggering this warning when you have a method setServiceLocator in your controller?

With a factory you do not need the Service Locator in your controller.

bgaillard:
Hi, we also encounter this problem, sorry to complain again but this is the third problem we have in one week with the ZF 2 updates :(  

Could this be fixed quickly ?

wtabata:
In all my controllers appear as message: " Deprecated: ServiceLocatorAwareInterface is deprecated and will be removed in version 3.0 [...]"

[...]

PR #88, opened 2016-03-02
Better ServiceLocatorAware deprecation

weierophinney:
This is a patch for fixing #87.

We were:
- injecting any service that _was not_ marked as `ServiceLocatorAware` but implemented `setServiceLocator()`, and
- injecting any service that _was_ marked `ServiceLocatorAware`, and
- emitting a deprecation notice in both cases.

The rationale was to allow backwards compatibility with existing implementations, while still messaging to users that they need to update their code not to rely on that.

This created a few problems, though:
- _Every_ controller inheriting from `AbstractController` now emits deprecation notices on instantiation.
- Developers who had implemented a `setServiceLocator()` method without implementing `ServiceLocatorAwareInterface` were now getting auto-injected with the service locator instance when they did not expect to.

This patch attempts to address those issues as follows:
- `ServiceManagerConfig` updates its service locator initializer to:
  - Inject non-plugin manager `ServiceLocatorAware` instances; in this situation, a deprecation notice _is_ emitted.
  - For `ServiceLocatorAware` plugin managers, it checks to see if a service locator is already composed. If not, it will inject it, but then emit a deprecation notice indicating that the service locator should be injected as a constructor parameter.
- `ControllerManager` updates its service locator initializer to:
  - For non-`ServiceLocatorAware` `AbstractController` implementations, it injects the service locator with no notices.
  - For all explicitly `ServiceLocatorAware` controllers, it injects the service locator _and_ emits a deprecation notice.
  - It no longer injects non-`ServiceLocatorAware` controllers that implement `setServiceLocator()` unless they extend `AbstractController`. This is fully backwards compatible.
- `AbstractController::getServiceLocator()` now emits a deprecation notice prior to returning the service locator instance. This is more appropriate, as we want to encourage users who rely on the service locator composition to upgrade.

Interestingly, this allows us to revert a number of changes that were made to the tests to mask deprecation notices, which helps solidify the argument that the original changes were indeed a backwards compatibility break, and that the new changes will address that situation.


froschdesign:
Better! :+1:

snapshotpl:
Much better! :+1:

snapshotpl:
I see one case which allow users to use service locator in controllers without warning - by serviceLocator property. Are we should care about this? @weierophinney

weierophinney:
@snapshotpl Too difficult to manage (we'd have to use property overloading, and that gets messy fast). Those users will feel the pain when they upgrade to zend-mvc v3 later, and it will be covered in the migration guide at that time.

bgaillard:
@weierophinney Thanks, it works on our project now and we do not encounter the deprecated warning :+1:

kevin-olbrich:
I encountered a problem regarding this deprecation:

Sample Code:

```
        $em = $this->getServiceLocator()->get('doctrine.entitymanager.orm_default');

        echo "test1".PHP_EOL;

        $machines = $em->getRepository('Main\Entity\DataMachines')->findAll();

        echo "test2".PHP_EOL;
        die("die here...");
```

Output:

> C:\secret>php public\index.php my-cli-job
> 
> Deprecated: You are retrieving the service locator from within the class CronJobs\Controller\SecretController. Please be aware that ServiceLocatorAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceLocatorAwareInitializer. You will need to update your class to accept all dependencies at creation, either via constructor arguments or setters, and use a factory to perform the injections. in C:\secret\vendor\zendframework\zend-mvc\src\Controller\AbstractController.php on line 258
> test1
> 
> C:\secret>

The job silently dies when I call a function of doctrine.

kevin-olbrich:
Problem fixed. Doctrine-only problem.
https://github.com/doctrine/DoctrineORMModule/issues/461

chrisVdd:
Hi!

Could you please explain me how I can have this patch in my local project please ?
Thanks!

froschdesign:
@chrisVdd 
Update the `Zend\Mvc` component per composer. For example:
- Version 2.7.10 for ZF2
- Version 3.0.3 for ZF3

Issue #92, opened 2016-03-07
using middleware in 'dispatch' event?


samsonasik:
is it possible to use it in a 'dispatch' event, tried something like this:

``` php
$eventManager->attach('dispatch', $sm->get('AuthorizationMiddleware'));
```

and got the following error:

```
: Missing argument 2 for Application\Middleware\AuthorizationMiddleware::__invoke() in /Application/Middleware/AuthorizationMiddleware.php on line 7
```

While the error is argument 2, the argument 1 is an `MvcEvent` instance.

weierophinney:
No, because they accept completely different arguments!

Event listeners receive an event instance, while middleware accepts a request, response, and optionally "next" callable. On top of that, the request and response from the MvcEvent are the zend-http types.

You could certainly wrap the middleware to make this happen:

``` php
use Zend\Psr7Bridge\Psr7Request;
use Zend\Psr7Bridge\Psr7Response;

$events->attach('dispatch', function ($e) use ($services) {
    $request = Psr7Request::fromZend($e->getRequest());
    $response = Psr7Response::fromZend($e->getResponse());
    $result = ($services->get('AuthorizationMiddleware'))($request, $response);
    // determine if a response should be returned, etc.
});
```

samsonasik:
Thank you... I think it need to be documented ;)

Warm regards,

Abdul Malik Ikhsan

Pada 7 Mar 2016, pukul 23.41, weierophinney notifications@github.com menulis:

weierophinney:
Why? The MiddlewareListener details its scope clearly already, and that scope does not include arbitrary event listeners... If you'd like to contribute a cookbook recipe, however, feel free to send a pull request. :wink:

samsonasik:
@weierophinney here we go https://github.com/zendframework/zend-mvc/pull/94 for the PR doc ;)

samsonasik:
closing in favor of #94

PR #94, opened 2016-03-07
add doc for middleware usage in "dispatch" event

samsonasik:
#92


weierophinney:
@samsonasik Please create a new `cookbook/` subtree in the documentation source, and add this. Then, in `mkdocs.yml`, create a `Cookbook` subsection, with this as the only sub-document.

Thanks!

samsonasik:
@weierophinney done ;). moved to cookbook and updated mkdocs.yml

doc/book/cookbook/using-middleware-in-dispatch-event.md
```diff
@@ -0,0 +1,41 @@
...
+    public function onBootstrap($e)
+    {
+        $app          = $e->getApplication();
+        $eventManager = $app->getEventManager();
+        $services     = $app->getServiceManager();
+
+        $eventManager->attach('dispatch', function ($e) use ($services) {
+            $request  = Psr7ServerRequest::fromZend($e->getRequest());
+            $response = Psr7Response::fromZend($e->getResponse());
+            $result   = ($services->get(AuthorizationMiddleware::class))($request, $response);
```
danizord: I think that you could pass an empty closure as third parameter to the middleware, so that if the middleware `return $next($req, $res)`, then $result will be `null`, and if the middleware `return $response`, then `$result` will be a response. In case it returns a response, you can convert it back to Zend\Http and return it (short-circuiting).

samsonasik: Done ;)

danizord:
@samsonasik now in listener, you can add this:

``` php
if ($result instanceof \Psr\Http\Message\ResponseInterface) {
    return Psr7Response::toZend($result);
}
```

So that the middleware becomes able to short-circuit, for example some `AuthorizationMiddleware` can return a 403 response if not authorized. :)

samsonasik:
@danizord done ;)  Thank you ;)

weierophinney:
Thanks, @samsonasik !

Issue #90, opened 2016-03-03
[2.7.1] Zend\Mvc\Controller\Plugin\Forward


ddetyuk:
Zend\Mvc\Controller\Plugin\Forward do not detach Zend\Mvc\View\Http\InjectViewModelListener listener

https://github.com/zendframework/zend-mvc/blob/master/src/Controller/Plugin/Forward.php#L195-L196

sould be

``` php
    $callbacks = $currentCallback->getCallback();
    if ($callbacks[0] instanceof $class) {
        $sharedEvents->detach($id, $currentCallback);
```

instead of

``` php
    if ($currentCallback instanceof $class) {
        $sharedEvents->detach($currentEvent, $id);
```

weierophinney:
Referenced the wrong issue in a commit I pushed; re-opening.

PR #96, opened 2016-03-08
Ensure listener detachment works cross-version

weierophinney:
Check if a returned callback is a CallbackHandler, and, if so, pull the actual PHP callable from it before testing for the class. To ensure both BC and forwards compatibility, ensure that the priority is stored when detaching, so that re-attaching will continue to work.

In order to use `composer update --prefer-lowest` (which I did for testing against zend-eventmanager v2), I had to:
- update PHPUnit to `^4.5`, as that's the first version using prophecy, which we're using in tests now.
- add sebastian/version:^1.0.4, as that's the first version that fixes version autodiscovery, ensuring that the test runner knows we're on a version later than 2.5.

Tests for this are ridiculously hard to write; would love some review!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,8 @@
         "zendframework/zend-version": "^2.5",
         "zendframework/zend-view": "^2.6.3",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/PHPUnit": "~4.0"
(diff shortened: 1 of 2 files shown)


weierophinney:
Forgot to note that this addresses #90.

PR #99, opened 2016-03-09
Consume zend-router and remove internal Router subcomponent

weierophinney:
Now that [zend-router](https://github.com/zendframework/zend-router) has been split out, we can remove it from the zend-mvc distribution, and add zend-router as a dependency.

Changes required:
- Removal of the `Zend\Mvc\Router` subcomponent and related tests.
- Removal of `Zend\Mvc\Service` factories that are now in zend-router.
- All references to the Router subcomponent were updated to reference the new namespace.
- Updated the various `Router` entries in the `ServiceListener` to refer to the factories from zend-router.
- To prevent issues with alias resolution, updated several tests to define a `Router` factory that simply returns the `HttpRouter` service.
- The PRG plugin test needed to change the name of one expected exception; it likely should have been updated sooner, regardless.
## TODO
- [x] remove the entry for the `ConsoleRouter`, as this will be optional functionality provided by zend-mvc-console, and will not need to be mapped in the `ServiceListener`.
- ~~[ ] zend-view needs to be updated to consume zend-router; as such we'll need to update this PR again in the future to reference the new zend-view version once it's tagged.~~ This can and should be merged before zend-view is updated; we'll just need to update once zend-view _is_ updated.
- [x] experiment with having the tests use the zend-router `Module` for providing service mappings, to see if it can be consumed seamlessly that way.
- [x] since console routes are removed from zend-router, and [zend-mvc-console now exists](https://zendframework.github.io/zend-mvc-console/), remove all console functionality.


weierophinney:
Labeling as a WIP, as I plan to get this working with zend-router v3.

PR #111, opened 2016-03-31
use default console exception template instead of empty string

svycka:
this was overriding default template if no template is set.

also the method name and config option name 'exception_message' is not very clear what it is :D

### DIFF
--- a/src/Service/ConsoleExceptionStrategyFactory.php
+++ b/src/Service/ConsoleExceptionStrategyFactory.php
@@ -68,7 +68,8 @@ class ConsoleExceptionStrategyFactory implements FactoryInterface
      */
     private function injectExceptionMessage(ExceptionStrategy $strategy, array $config)
     {
-        $message = isset($config['exception_message']) ? $config['exception_message'] : '';
-        $strategy->setMessage($message);
+        if (isset($config['exception_message'])) {
+            $strategy->setMessage($config['exception_message']);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


svycka:
@weierophinney I think this was introduced after your SM v3 refactor and merged into 2.7 is this desired behavior or bug? If not a bug then why we have default template and overriding anyway.

snapshotpl:
@svycka can you add unit test to cover this bug?

svycka:
in case this will not merged soon as a workaround I added default template from here https://github.com/zendframework/zend-mvc/blob/master/src/View/Console/ExceptionStrategy.php#L31-L44 to my local config like this:

```
return [
    'console' => [
        'view_manager' => [
            'exception_message' => '
======================================================================
   The application has thrown an exception!
======================================================================
 :className
 :message
----------------------------------------------------------------------
:file::line
:stack
======================================================================
   Previous Exception(s):
:previous

',
        ]
    ]
];
```

> @svycka can you add unit test to cover this bug?

@snapshotpl not sure but maybe if really needed

svycka:
@snapshotpl is it ok?

svycka:
@weierophinney this bug also is a BC break and maybe can be fixed faster?

weierophinney:
@svycka Merged, and tagged 2.7.5. Can you do a similar PR against zend-mvc-console, please?

svycka:
@weierophinney done https://github.com/zendframework/zend-mvc-console/pull/5

PR #112, opened 2016-03-31
Fix for Forward in zend-eventmanager v3

rwrz:
Actually,

EventManager returns an ARRAY on `$sharedEvents->getListeners` and the way it was implemented the `listenersToDetach` it wasn't working. 
I don't know if in V2 it is the same thing, so, this pull request FIX this method and a problem from rendering the View two times.

Like this old bug: http://zend-framework-community.634137.n4.nabble.com/Double-rendering-with-forward-plugin-td4656067i20.html

Someone can check if it works on `zend-eventmanager` v2?

Thanks,

Rodrigo

### DIFF
--- a/src/Controller/Plugin/Forward.php
+++ b/src/Controller/Plugin/Forward.php
@@ -184,6 +184,7 @@ class Forward extends AbstractPlugin
                     if (!is_array($currentPriorityEvents)) {
                         $currentPriorityEvents = array($currentPriorityEvents);
                     }
+                    
                     // v3
                     foreach ($currentPriorityEvents as $currentEvent) {
                         $currentCallback = $currentEvent;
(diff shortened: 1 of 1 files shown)


weierophinney:
@rwrz Travis tests against lowest supported, current lockfile, and latest dependencies; the fact you have a green light from Travis indicates it works against zend-eventmanager v2 as well. :smile:

src/Controller/Plugin/Forward.php
```diff
@@ -179,32 +179,39 @@ protected function detachProblemListeners(SharedEvents $sharedEvents)
...
-                    // zend-eventmanager v2 compatibility:
-                    if ($currentCallback instanceof CallbackHandler) {
-                        $currentCallback = $currentEvent->getCallback();
-                        $priority        = $currentEvent->getMetadatum('priority');
+                foreach ($events as $priority => $currentPriorityEvents) {
+                    // v2 fix
+                    if (!is_array($currentPriorityEvents)) {
+                        $currentPriorityEvents = [$currentPriorityEvents];
                     }
+                    // v3
```
weierophinney, line 187: This comment can be removed, as the logic below executes for either version. (I'll do that during merge.)

PR #116, opened 2016-04-05
Remove zend-di <-> zend-servicemanager integration

weierophinney:
zend-di <-> zend-servicemanager integration is now provided by [zend-servicemanager-di](https://zendframework.github.io/zend-servicemanager-di/), and can be safely removed from zend-mvc.

This patch:
- Removes the functionality exposed previously by zend-mvc, including related tests.
- Removes the zend-di dependency from `composer.json`, and adds a suggestion for zend-servicemanager-di.
- Provides migration documentation for end-users.


Ocramius:
This looks :+1: from here!

PR #126, opened 2016-04-12
Do not throw deprecated notice if serviceLocator is set

svycka:
I am not sure this is desired behavior, but seems strange. Noticed this while reviewing zfcUser module that removing `ServiceManagerAwareInterface` would be BC break. See this comment https://github.com/ZF-Commons/ZfcUser/pull/634#discussion-diff-59324273

For example for AbstractPluginManager you added check if service locator is set then do not throw deprecated error, but for modules like zfcUser there is no way to make it BC way.

@weierophinney maybe I am missing something?

### DIFF
--- a/src/Service/ServiceManagerConfig.php
+++ b/src/Service/ServiceManagerConfig.php
@@ -150,6 +150,7 @@ class ServiceManagerConfig extends Config
                 // implementations; they're dealt with later.
                 if ($instance instanceof ServiceLocatorAwareInterface
                     && ! $instance instanceof AbstractPluginManager
+                    && ! $instance->getServiceLocator()
                 ) {
                     trigger_error(sprintf(
                         'ServiceLocatorAwareInterface is deprecated and will be removed in version 3.0, along '


src/Service/ServiceManagerConfig.php
```diff
@@ -150,6 +150,7 @@ public function __construct(array $config = [])
                 // implementations; they're dealt with later.
                 if ($instance instanceof ServiceLocatorAwareInterface
                     && ! $instance instanceof AbstractPluginManager
+                    && ! $instance->getServiceLocator()
```
Ocramius, line 153: Is this the original behavior? Can't remember if we did overwrite the `ServiceLocatorInterface` instance each time or not (see previous framework versions). We can't change this, if we did overwrite it at each occasion.

svycka, line 153: @Ocramius I think yes old versions always set servicelocator. 

``` php
'ServiceLocatorAwareInitializer' => function ($instance, ServiceLocatorInterface $serviceLocator) {
    if ($instance instanceof ServiceLocatorAwareInterface) {
        $instance->setServiceLocator($serviceLocator);
    }
},
```

But what it change? Do you think this would mean BC break? If someone has BC issues with this would be really bad code. Also AbstractPluginManager then have BC break it does not set service locator if already set in current version?

Anyway if we can't change this then how we suppose update for example zfcUser backwards compatible way?

weierophinney, line 153: The changes in the 2.7 branch are for the following reasons:
- We test for plugin managers, because we want to vary the deprecation message: plugin managers need the parent locator passed during instantiation in order to be forwards compatible, while non-plugin manager, ServiceLocatorAware instances should be updated to have their dependencies injected discretely.
- The original behavior was, as noted, to always inject, without regard for whether or not an instance was present. This is _still_ true, except in the case of abstract plugin managers. There is no BC break.

The reason for the change with regard to abstract plugin managers is that all of those managed by ZF itself have been updated at this point to inject the parent locator during instantiation, and we do not want to raise the deprecation notice for them, as they're already updated. The new code block is literally for _custom_ plugin managers that are registered by end-users, to notify them they need to update their factories. In fact, it was introduced because they were _not_ getting injected by previous iterations in the 2.7 series, which was causing errors.

@svycka I guess what I'm wondering is: what exact messages are you getting from ZfcUser? Can you share the message and stack trace? Are there any custom plugin managers in ZfcUser? If so, where are the factories for them?

weierophinney:
Closing, due to lack of feedback from the author.

PR #139, opened 2016-05-17
[BC break] Remove old InjectTemplateListener behavior

Xerkus:
Removed old pseudomodule template name resolution which ignored subnamespaces and used `__NAMESPACE__` variable in convoluted and non-intuitive ways. 
Whitelisted classname based template name resolution introduced in `^2.3.0` is now default
Template resolution follows these rules:
1. strip `\Controller\` namespace if present
2. strip trailing `Controller` in classname, unless classname is exactly `Controller`
3. inflect CamelCase to dash
4. replace namespace separator with slash

See https://github.com/zendframework/zendframework/pull/5670

This breaks backwards compatibility in some cases, check couple changed tests.

Affected modules can achieve compatibility with v2 and v3 simultaneously by adding whitelist entry to template map and moving templates or by mapping namespace to old location.

Needs documentation. If anyone is willing to write it for me I definitely won't object to that

### DIFF
--- a/src/View/Http/InjectTemplateListener.php
+++ b/src/View/Http/InjectTemplateListener.php
@@ -12,7 +12,6 @@ namespace Zend\Mvc\View\Http;
 use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface as Events;
 use Zend\Mvc\MvcEvent;
-use Zend\Mvc\ModuleRouteListener;
 use Zend\Stdlib\StringUtils;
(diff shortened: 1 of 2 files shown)


weierophinney:
@Xerkus Could you please add some verbiage into `doc/book/migration/to-v3-0.md` detailing the impact of this for users? I can then merge.

Thanks!

weierophinney:
@Xerkus You can even leave a comment on the issue if you want, and I can then draft up the migration doc changes.

Issue #103, opened 2016-03-17
Deprecated error. Need a quick solution please!


xLeonius:
I have been searching for a solution to this issue for 3 days and I couldn't find an appropriate solution. Every time I get the same Issue, it just says deprecated. Here is what I get on my website:

Deprecated: ServiceManagerAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceManagerAwareInitializer. Please update your class SmallUser\Service\User to remove the implementation, and start injecting your dependencies via factory instead. in C:\Apache24\htdocs\pserverCMSFull\vendor\zendframework\zend-mvc\src\Service\ServiceManagerConfig.php on line 123

Deprecated: You are retrieving the service locator from within the class PServerCore\Controller\IndexController. Please be aware that ServiceLocatorAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceLocatorAwareInitializer. You will need to update your class to accept all dependencies at creation, either via constructor arguments or setters, and use a factory to perform the injections. in C:\Apache24\htdocs\pserverCMSFull\vendor\zendframework\zend-mvc\src\Controller\AbstractController.php on line 258

Deprecated: ServiceManagerAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceManagerAwareInitializer. Please update your class PServerCore\Service\News to remove the implementation, and start injecting your dependencies via factory instead. in C:\Apache24\htdocs\pserverCMSFull\vendor\zendframework\zend-mvc\src\Service\ServiceManagerConfig.php on line 123

Deprecated: ServiceManagerAwareInterface is deprecated and will be removed in version 3.0, along with the ServiceManagerAwareInitializer. Please update your class PServerCore\Service\ConfigRead to remove the implementation, and start injecting your dependencies via factory instead. in C:\Apache24\htdocs\pserverCMSFull\vendor\zendframework\zend-mvc\src\Service\ServiceManagerConfig.php on line 123
<
What I tried:
- Tried to remove the error reporting part from the .php files which cause these errors.
- Tried to re-download the whole zend-mvc folder from github, didn't work.

Please keep in mind that I am not an expert in such things, so please give me detailed steps to fix this issue. I'd really be grateful ;)

weierophinney:
[...]

PR #143, opened 2016-05-31
Added section on updating controllers to avoid deprecation notices

weierophinney:
Added material from [a comment I made](https://github.com/zendframework/zend-mvc/issues/103#issuecomment-198018642) regarding the new deprecation notices, and how to update your code to prevent them.

### DIFF
--- a/doc/book/migration.md
+++ b/doc/book/migration.md
@@ -103,3 +103,100 @@ injected by one of these initializers, and we plan to remove the initializers
 for version 3.0. The deprecation notice includes the name of the class, to help
 you identify what instances you will need to update before the zend-mvc v3
 release.
+
+To prepare your code, you will need to do the following within your controller:
+
+- Find all cases where you call `getServiceLocator()`, and identify the services
+  they retrieve.
+- Update your controller to accept these services via the constructor.
+- If you have not already, create a factory class for your controller.
+- In the factory, pull the appropriate services and pass them to the
+  controller's constructor.
+
+As an example, consider the following code from a controller:
+
+```php
+$db = $this->getServiceLcoator()->get('Db\ApplicationAdapter');
+```
+
+To update your controller, you will:
+
+- Add a `$db` property to your class.
+- Update the constructor to accept the database adapter and assign it to the
+  `$db` property.
+- Change the above line to either read `$db = $this->db;` *or just use the
+  property directly*.
+- Add a factory that pulls the service and pushes it into the controller.
+
+The controller then might look like the following:
+
+```php
+use Zend\Db\Adapter\AdapterInterface;
+use Zend\Mvc\Controller\AbstractActionController;
+
+class YourController extends AbstractActionController
+{
+    private $db;
+
+    public function __construct(AdapterInterface $db)
+    {
+        $this->db = $db;
+    }
+
+    public function someAction()
+    {
+        $results = $this->db->query(/* ... */);
+        /* ... */
+    }
+}
+```
+
+A factory would look like the following:
+
+```php
+use Interop\Container\ContainerInterface;
+
+class YourControllerFactory
+{
+    public function __invoke(ContainerInterface $container)
+    {
+        return new YourController($this->get('Db\ApplicationAdapter'));
+    }
+}
+```
+
+You then also need to ensure the controller manager knows about the factory. It
+likely already does, as an invokable; you will redefine it as a factory in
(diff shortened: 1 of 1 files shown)


doc/book/migration.md
```diff
@@ -103,3 +103,100 @@ injected by one of these initializers, and we plan to remove the initializers
...
+A factory would look like the following:
+
+```php
+use Interop\Container\ContainerInterface;
+
+class YourControllerFactory
+{
+    public function __invoke(ContainerInterface $container)
+    {
+        return new YourController($this->get('Db\ApplicationAdapter'));
```
snapshotpl, line 163: `$container` not `$this` ;-)

ghost:
I'm starting a new project and encountered this error.

As there will be loads of controllers and methods I don't want to have misunderstood anything and incorrectly start this project.

According to what I read on the web and your solution I undestand that we should create a factory per controller. Is that correct? 

I previously used a single method to get the Doctrine EntityManager and called it whenever I needed it. I used:

`$this->getServiceLocator()->get('Doctrine\ORM\EntityManager')`

This method now triggers a deprecated warning. Does it mean that I should move it to a service and pass it to every controller through specific factories?

As I might have more than 100 controllers this seems to be quite redundant and long to develop... Would there be any better and clean solution?

froschdesign:
@aspic10 

> Would there be any better and clean solution? 

Sure, use an [abstract factory](https://zendframework.github.io/zend-servicemanager/configuring-the-service-manager/#abstract-factories).

ghost:
Thanks! I had never used abstract factories and it seems to be a simple and clean solution. Again, to be sure I start correctly I need your comments on what I did.

I created an abstract factory that will instantiate the controllers and pass the EntityManager as an argument.

In the module.config.php I removed the controllers from the "invokables" and "factories" sections and added my abstract factory in the "abstract_factories" section.

If a controller needs different arguments I will add it to the "factories" and create a specific factory.

Is this correct?

michalbundyra:
@aspic10 Yes, it's correct. Service Manager will use factory at first if it is given.
If there is no factory for your service then SN goes through all abstract factories.
Please read secion about [Factories in documentation](https://zendframework.github.io/zend-servicemanager/configuring-the-service-manager/#factories). Everything is there explained with some important "Best practices"

froschdesign:
> Everything is there explained with some important "Best practices"

Right!

ghost:
Thank you very much. I understand factories much better now and will improve my code to make it cleaner and more performant.

weierophinney:
One quick note: with zend-servicemanager v3 (which MVC v3 requires), abstract factories are also compatible for use as factories. This allows you to map multiple controller names to the same factory, making it more explicit how they will be created.

ghost:
From what I read in the link provided by webimpress and weierophinney's answer the best practice would be to use a common factory instead of an abstract factory.

I'm still not sure yet that I have got it all and that I'm doing it right though. Is this example correct:

In module.config.php:

```
'controllers' => array(
    'factories' => array(
        Controller\Index::class => Factory\DefaultControllerFactory::class,
        Controller\Other::class => Factory\DefaultControllerFactory::class,
        Controller\Third::class => Factory\DefaultControllerFactory::class,
        Controller\User::class => Factory\UserFactory::class,
    ),
),
```

And in the default factory:

```
public function __invoke($container, $name, $requestedName, array $options = null) {
    $serviceManager = $container->getServiceLocator();
    $entityManager = $serviceManager->get('doctrine.entitymanager.orm_default');

    $controller = $requestedName . "Controller";
    if(class_exists($controller)){
        return new $controller($entityManager);
    }
    else {
        return false;
    }
}
```

I had previously defined an abstract factory that implemented AbstractFactoryInterface and did pretty much the same, except controllers related to it weren't defined under "invokables" nor "factories" in the module.config.php file.

michalbundyra:
@aspic10 you can use abstract factory if you'd prefer, it would be nothing bad with it, and as you said then you don't need define "invokable" nor "factories" in config. One abstract factory is not going make your application dramatically less efficient :) Please notice that this abstract factory will be defined only for controllers (controller manager) - so it will be checked **only** on creating controller.

If you decide use common factory, then you need define these relations in `module.config.php` as you did. In your default factory code probably `$requestedName` is already controller name, so you don't need append "Controller" and you don't need return `false` if controller is not found.

Personally, I would go with abstract factory in that case.

Issue #148, opened 2016-06-02
Notice: Undefined variable: serviceLocator


simonmacor:
Hi,
The error is in Service/DiAbstractServiceFactoryFactory.php on line 32
I'm getting this error when I try create a service with a factory.

```
public function __invoke(ContainerInterface $container, $name, array $options = null)
    {
        $factory = new DiAbstractServiceFactory($container->get('Di'), DiAbstractServiceFactory::USE_SL_BEFORE_DI);

        if ($serviceLocator instanceof ServiceManager) {
            $serviceLocator->addAbstractFactory($factory, false);
        }

        return $factory;
    }
```

The problem seems to be present since the release 2.7.0

michalbundyra:
@weierophinney could you please create branch `release-2.7` and then will be possible to create PR with the fix against this branch?

weierophinney:
@webimpress Branch is now created, based on release-2.7.8. Ping me when you have a PR ready.

weierophinney:
Fixed with #158.

PR #149, opened 2016-06-02
Notice: Undefined variable: serviceLocator

michalbundyra:
@weierophinney here it is: fix for #148. I've added some tests, not sure if they are useful because as I can see this part is removed from v3. Let me know please what do you think. Thanks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "php": "^5.5 || ^7.0",
         "zendframework/zend-eventmanager": "^2.6.2 || ^3.0",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
+        "zendframework/zend-servicemanager-di": "^1.0",
         "zendframework/zend-hydrator": "^1.1 || ^2.1",
         "zendframework/zend-form": "^2.7",
         "zendframework/zend-stdlib": "^2.7.5 || ^3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "1ead671a9764c969d80511a14f87dd0b",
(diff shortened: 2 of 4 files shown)


weierophinney:
Okay, I see a couple of things.

First, as you note, the functionality is removed in zend-servicemanager v3, which means the tests fail if that version is installed.

Interestingly, zend-servicemanager-di, which is where we moved the functionality for the zend-mvc v3 release, _is_ tested against both versions, and contains the exact same functionality, albeit with different namespaces. Unfortunately, it contains the same typo. :frowning: I'll be doing a PR for that momentarily so we can get a working version in place.

My thought is that we do the following:
- Modify `DiAbstractServiceFactoryFactory` to extend `Zend\ServiceManager\Di\DiAbstractServiceFactoryFactory`
- Do the same for `DiAbstractServiceFactory` (have it extend the zend-servicemanager-di variant)
- Add a requirement in `composer.json` for zend-servicemanager-di

The above will solve the issue so that zend-mvc v2 continues to work, both against zend-servicemanager v2 and v3.

michalbundyra:
@weierophinney I've updated my PR, please have a look. Thanks.

weierophinney:
@webimpress I was working on similar changes, after reviewing this and zend-servicemanager-di.

I took your original changeset and was able to apply it against zend-servicemanager-di (with attribution, even; I used the `.patch` from this PR, edited it, and applied it to that component, which gives you credit for the commit), which fixed the related issue in that component. I've now opened #158, which:
- adds zend-servicemanager-di to the component requirements
- updates 4 of the DI-related factories to simply extend from the zend-servicemanager-di variants
- marks all 5 of the DI-related factories as deprecated via annotation

This approach is completely BC; the four I created as extensions were returning artifacts from the `Zend\ServiceManager\Di` namespace already, and the fifth now returns one of these extending classes, preserving inheritance.

Thanks for raising the issue, and for your patches!

michalbundyra:
@weierophinney thank you, it's much better now, thanks :)

PR #150, opened 2016-06-02
consistent names for exception (v3)

michalbundyra:
small fix - renamed variables (for consistency with other occurrences) and to avoid notice in IDE "variable might have not been defined"


weierophinney:
Thanks, @webimpress !

Issue #146, opened 2016-06-01
Throwable vs. Exception


PhoneBuster:
I use zend-mvc 2.7.8. If i have a bad argument in my own factory the framework throws the following error :

**Fatal error: Uncaught TypeError: Argument 4 passed to Zend\Mvc\DispatchListener::marshalBadControllerEvent() must be an instance of Exception, instance of TypeError given...**

To fix this i've to change the the argument type in the DispatchListener class (method marshalBadControllerEvent) from 

``` php
   protected function marshalBadControllerEvent(
        $controllerName,
        MvcEvent $event,
        Application $application,
        \Exception $exception
    ) {
        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
        $event->setError($application::ERROR_EXCEPTION);
        $event->setController($controllerName);
        $event->setParam('exception', $exception);

        $events  = $application->getEventManager();
        $results = $events->triggerEvent($event);
        $return  = $results->last();
        if (! $return) {
            return $event->getResult();
        }

        return $return;
    }
```

to

``` php
   protected function marshalBadControllerEvent(
        $controllerName,
        MvcEvent $event,
        Application $application,
        Throwable $exception
    ) {
        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
        $event->setError($application::ERROR_EXCEPTION);
        $event->setController($controllerName);
        $event->setParam('exception', $exception);

        $events  = $application->getEventManager();
        $results = $events->triggerEvent($event);
        $return  = $results->last();
        if (! $return) {
            return $event->getResult();
        }

        return $return;
    }
```

Ocramius:
Due to PHP 5.x compatibility, the quickest solution for now is to drop the type-hint completely

PhoneBuster:
My PHP Version is 7

Ocramius:
The library still supports some versions of PHP 5.x

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

weierophinney:
Fixed with #153 and release 2.7.9. You may also need to update your error templates per the suggestions in #157.

andreaventuri:
I still have the same problem in version 3.0.1

michalbundyra:
It was fixed only on v2.7, here you have issue for v3: https://github.com/zendframework/zend-mvc/pull/161

/cc @weierophinney

weierophinney:
Fixed with #161.

PR #153, opened 2016-06-03
call with Throwable, but type-hint wants Exception

kokspflanze:
hi,

i get the same problem like https://github.com/zendframework/zend-mvc/issues/146 with release-3.0.0.

in `Zend\Mvc\DispatchListener` https://github.com/zendframework/zend-mvc/blob/master/src/DispatchListener.php#L103 is the method call with `Throwable` but the type-hint want an `Exception` class https://github.com/zendframework/zend-mvc/blob/release-3.0.0/src/DispatchListener.php#L219

Im not sure if the fix should be only in the release-3.x.x. or also in release-2.7.x.

PHP-Version 7.0.7

### DIFF
--- a/src/DispatchListener.php
+++ b/src/DispatchListener.php
@@ -235,14 +235,14 @@ class DispatchListener extends AbstractListenerAggregate
      * @param  string $controllerName
      * @param  MvcEvent $event
      * @param  Application $application
-     * @param  \Exception $exception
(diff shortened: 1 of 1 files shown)


sandrokeil:
I've the same problem.

andreaventuri:
I still have the same problem in version 3.0.1

PR #155, opened 2016-06-06
Update docs to reflect template map resolver change in v3

Xerkus:


doc/book/migration/to-v3-0.md
```diff
@@ -412,22 +412,22 @@ the section above on [dependency reduction](#dependency-reduction).
...
+had a controller with the same name.
 
 To avoid naming conflicts, version 3 removes this aspect of autodetection, and
 instead provides exactly one workflow for mapping:
 
 - Strip the `Controller` subnamespace, if present (e.g.,
-  the namespace `Application\Controller\\` is normalized to
-  `Application\\`).
+  the namespace `Application\Controller\` is normalized to
+  `Application\`).
```
weierophinney, line 430: The double-backslashes were actually required to ensure that the backtick isn't considered a literal. :smile: Re-instating during merge.

Xerkus, line 430: @weierophinney I noticed that github markdown renderer does not treat backslash as special character in backticked strings and docs site appears to be using same renderer:
https://zendframework.github.io/zend-mvc/migration/to-v3-0/#zendmvcviewinjecttemplatelistener

weierophinney, line 430: I'll double check. Syntastic was flashing it in my editor, and I've run
into issues before.

PR #159, opened 2016-06-13
Better zend-servicemanager <-> zend-di integration deprecation

weierophinney:
Per #125, requiring zend-servicemanager-di with zend-servicemanager v2 causes conflicts when creating an optimized autoloader with Composer. Unfortunately, the issue we're left with is:
- zend-servicemanager v3 removes the `Zend\ServiceManager\Di` namespace, which makes the various DI-related factories in zend-mvc result in fatal errors due to missing classes.
- unfortunately, there's no ability to do conditional requires in Composer, so we cannot specify a rule like "if zend-servicemanager v3 is installed with zend-mvc v2, install zend-servicemanager-di as well".

The only viable solution is to re-instate the code removed in 2.7.9, but with the following changes:
- the `DiAbstractServiceFactoryFactory` now checks to see if `Zend\ServiceManager\Di\DiAbstractServiceFactory` exists, and, if not, raises an exception indicating the developer should install zend-servicemanager-di.
- the `DiServiceInitializerFactory` now checks to see if `Zend\ServiceManager\Di\DiServiceInitializer` exists, and, if not, raises an exception indicating the developer should install zend-servicemanager-di.

Both of the above changes have unit tests for the new behavior introduced.

Additionally, all classes changed in 2.7.9 retain their deprecation annotations, signaling the change to come with zend-mvc v3.


akrabat:
Visual inspection of code changes look good to me.

PR #160, opened 2016-06-14
Add new params function "fromAny"

SilverioMiranda:
I use this function in my work to avoid rewriting code, across all projects we have like 5000+ inputs and making ifs for fromQuery,fromPost,fromRoute is bad.
This function will help if you dont need to know about if your param is from post,query or route


weierophinney:
Thanks for the contribution, @SilverioMiranda ! Unfortunately, this is not one we are inclined to accept, as it essentially implements `$_REQUEST` with the semantics of `variables_order`, which has long been a security issue within the PHP community. We recommend always considering the specific source of input when retrieving request parameters.

PR #161, opened 2016-06-17
marshalBadControllerEvent argument type hinting

autowp:
Exception -> Throwable

At line 102 marshalBadControllerEvent used with Throwable argument type

### DIFF
--- a/src/DispatchListener.php
+++ b/src/DispatchListener.php
@@ -216,7 +216,7 @@ class DispatchListener extends AbstractListenerAggregate
         $controllerName,
         MvcEvent $event,
         Application $application,
-        \Exception $exception
+        \Throwable $exception
     ) {
         $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
         $event->setError($application::ERROR_EXCEPTION);


src/DispatchListener.php
```diff
@@ -216,7 +216,7 @@ protected function marshalBadControllerEvent(
         $controllerName,
         MvcEvent $event,
         Application $application,
-        \Exception $exception
+        \Throwable $exception
```
michalbundyra, line 219: it is not gonna work on php 5.6 :(
please have a look in [`composer.json`](https://github.com/zendframework/zend-mvc/blob/master/composer.json#L16), and it still must work on: `"php": "^5.6 || ^7.0"`

michalbundyra:
it was already resolved in release [2.7.9](https://github.com/zendframework/zend-mvc/releases/tag/release-2.7.9):
https://github.com/zendframework/zend-mvc/pull/153/files

should be also ported to release v3

/cc @weierophinney

src/DispatchListener.php
weierophinney, line 219: @webimpress It works fine on PHP 5.6; the class does not need to exist for a typehint check.

weierophinney, line 219: Missed that this was in the method signature and not the docblock. I'll remove the typehint during merge.

weierophinney:
Thanks @autowp !

PR #163, opened 2016-06-19
InjectTemplateListener - returnn api proper view 

karborator:
- InjectTemplateListener  - `injectTemplate` : set `$preferRouteMatchController` view config param if exists. This way apigility rest service can return proper view per api .
- prefer_route_match_controller should be set at the config route to true  :
  
  ``` php
  'route' => '/sales[/:sales_id]',
                  'defaults' => array(
                      'controller' => 'Ecommerce\\V1\\Rest\\Sales\\Controller',
                      'prefer_route_match_controller'=>true
                  ),
  ```
- `controller_map` should be set at the config `view_manager` also:
  
  ``` php
    'controller_map' => array(
          'Ecommerce\\V1\\Rest\\Sales\\Controller' => 'ecommerce/sales'
    ),
  ```
- AcceptableViewModelSelector - notice fix

Issue #3, opened 2015-05-26
Unable to render template RuntimeException for AbstractRestfulController


jerv13:
I get an unable to render template error when i do not over-ride the AbstractRestfulController method.

By default, the response code is set and an array is returned which appears to be picked up and rendered using the default render strategy. 

Would it be a good idea to set the response content and return the response or is there something I am missing in the configuration.

I would just like some feedback on this before I make a pull request.

Thanks, 

James

rodmcnew:
I've ran into this issue many times as well and haven't found a good solution.

One way to recreate this is to extend the AbstractRestfulController and implement the get() action but not the getList() action. Then when you go to the getList route, you get a "template not found" error instead of a "Method Not Allowed" response like any other framework gives.

It doesn't make sense for me to have to create a "module-name/controller-name/get-list.phtml" template for every rest controller I create when I don't even want to implement "getList" functionality.

scroach:
I've stumbled across the same problem as I am currently trying to implement a REST API with ZF2.
I registered the `ViewJsonStrategy` in my module and expected the `AbstractRestfulController` to return JSON-Based error-responses for unimplemented methods.

Afte some research I found out that the `JsonStrategy` has been changed long ago in 2.0.4 to only return JSON if a JsonModel has been explicitely returned.
http://framework.zend.com/blog/zend-framework-2-0-4-released.html

So in my understanding the `AbstractRestfulController` shouldn't return plain arrays in the pre-implemented methods but `JsonModel` containing the array. This would produce something like this:

```
{"content":"Method Not Allowed"}
```

Am I right?!

karborator:
https://github.com/karborator/zend-mvc

chaos0815:
Is there a reason why AbstractRestfulController returns an array in the pre-implemented methods instead of an JsonModel object?

michalbundyra:
This repository has been closed and moved to laminas/laminas-mvc; a new issue has been opened at https://github.com/laminas/laminas-mvc/issues/34.

PR #164, opened 2016-06-19
InjectTemplateListener return proper api template

karborator:
InjectTemplateListener return proper api template if prefer_route_match_controller is set to true from route config.This way we can set controller_map and as a result ,apigility will return proper api template ,instead to look for template zf/rest/action.phtml ( get-list.phtml ) 

```
                'defaults' => array(
                    'controller' => 'Ecommerce\\V1\\Rest\\Sales\\Controller',
                    'prefer_route_match_controller' => true
                ),
```

### DIFF
--- a/src/View/Http/InjectTemplateListener.php
+++ b/src/View/Http/InjectTemplateListener.php
@@ -69,7 +69,7 @@ class InjectTemplateListener extends AbstractListenerAggregate
         }
 
         $routeMatch = $e->getRouteMatch();
-        if($preferRouteMatchController = $routeMatch->getParam('prefer_route_match_controller', false)) {
+        if ($preferRouteMatchController = $routeMatch->getParam('prefer_route_match_controller', false)) {
             $this->setPreferRouteMatchController($preferRouteMatchController);
         }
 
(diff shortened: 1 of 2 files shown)


weierophinney:
Please provide a test; i don't fully interstage what the patch is trying to accomplish.

karborator:
This patch is to solve issues like : https://github.com/zendframework/zend-mvc/issues/3

rodmcnew:
Just to clarify, #3 is about how the the AbstractRestfulController looks for templates and returns "500 internal server error" for undefined methods rather than returning "405 method not allowed".

Does this fix that?

karborator:
It will fix the problem with returning of proper template ... . when using rest API controller  
You can simple make a test .Clone my fork https://github.com/karborator/zend-mvc , 
create rest api , create controller_map of controller and template path and set  'prefer_route_match_controller' => true at the controller route .
Configure apigility  to return proper content negotiation also ... 

This way the inject listener will detect what template you want for your API 

From comments -  https://github.com/zendframework/zend-mvc/issues/3

 " I've ran into this issue many times as well and haven't found a good solution.

One way to recreate this is to extend the AbstractRestfulController and implement the get() action but not the getList() action. Then when you go to the getList route, you get a "template not found" error instead of a "Method Not Allowed" response like any other framework gives.

It doesn't make sense for me to have to create a "module-name/controller-name/get-list.phtml" template for every rest controller I create when I don't even want to implement "getList" functionality.   "

weierophinney:
Closing in favor of #163, which also incorporates the same commits.

PR #165, opened 2016-06-23
New LazyControllerFactory

weierophinney:
Inspired by http://circlical.com/blog/2016/3/9/preparing-for-zend-f, this abstract factory/factory can be used to create controller instances for controllers defining constructor dependencies, using the following rules:
- A parameter named `$config` typehinted as an array will receive the application "config" service (i.e., the merged configuration).
- Parameters type-hinted against array, but not named `$config` will be injected with an empty array.
- Scalar parameters will be resolved as null values.
- If a service cannot be found for a given typehint, the factory will raise an exception detailing this.
- Some services provided by Zend Framework components do not have entries based on their class name (for historical reasons); the factory contains a map of these class/interface names to the corresponding service name to allow them to resolve.

`$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.

As it implements zend-servicemanager's v3 AbstractFactoryInterface, it may be used as either an abstract factory, or by mapping controller class names to the factory.

### DIFF
--- a/src/Controller/LazyControllerAbstractFactory.php
+++ b/src/Controller/LazyControllerAbstractFactory.php
@@ -36,7 +36,7 @@ use Zend\Validator\ValidatorPluginManager;
  * <code>
  * 'controllers' => [
  *     'abstract_factories' => [
- *         LazyControllerFactory::class,
+ *         LazyControllerAbstractFactory::class,
  *     ],
  * ],
  * </code>
@@ -46,7 +46,7 @@ use Zend\Validator\ValidatorPluginManager;
  * <code>
  * 'controllers' => [
  *     'factories' => [
- *         MyControllerWithDependencies::class => LazyControllerFactory::class,
+ *         MyControllerWithDependencies::class => LazyControllerAbstractFactory::class,
  *     ],
  * ],
  * </code>
@@ -70,7 +70,7 @@ use Zend\Validator\ValidatorPluginManager;
  * `$options` passed to the factory are ignored in all cases, as we cannot
  * make assumptions about which argument(s) they might replace.
  */
-class LazyControllerFactory implements AbstractFactoryInterface
+class LazyControllerAbstractFactory implements AbstractFactoryInterface
 {
     /**
      * Maps known classes/interfaces to the service that provides them; only
--- a/src/Controller/LazyControllerAbstractFactory.php
+++ b/src/Controller/LazyControllerAbstractFactory.php
@@ -77,9 +77,11 @@ class LazyControllerAbstractFactory implements AbstractFactoryInterface
      * required for those services with no entry based on the class/interface
      * name.
      *
+     * Extend the class if you wish to add to the list.
+     *
      * @var string[]
      */
-    private $aliases = [
+    protected $aliases = [
         ConsoleAdapterInterface::class  => 'ConsoleAdapter',
         FilterPluginManager::class      => 'FilterManager',
         HydratorPluginManager::class    => 'HydratorManager',
--- a/src/Controller/LazyControllerAbstractFactory.php
+++ b/src/Controller/LazyControllerAbstractFactory.php
@@ -114,9 +114,10 @@ class LazyControllerAbstractFactory implements AbstractFactoryInterface
             return new $requestedName();
         }
 
-        $parameters = array_map(function (ReflectionParameter $parameter) use ($container, $requestedName) {
-            return $this->resolveParameter($parameter, $container, $requestedName);
-        }, $reflectionParameters);
+        $parameters = array_map(
+            $this->resolveParameter($container, $requestedName),
+            $reflectionParameters
+        );
 
         return new $requestedName(...$parameters);
     }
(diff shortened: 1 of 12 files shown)


src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+{
+    /**
+     * Maps known classes/interfaces to the service that provides them; only
+     * required for those services with no entry based on the class/interface
+     * name.
+     *
+     * @var string[]
+     */
+    private $aliases = [
+        'Zend\Console\Adapter\AdapterInterface'     => 'ConsoleAdapter',
```
Ocramius: `::class` all these, plz

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+ *   raise an exception detailing this.
+ * - Some services provided by Zend Framework components do not have
+ *   entries based on their class name (for historical reasons); the
+ *   factory contains a map of these class/interface names to the
+ *   corresponding service name to allow them to resolve.
+ *
+ * `$options` passed to the factory are ignored in all cases, as we cannot
+ * make assumptions about which argument(s) they might replace.
+ */
+class LazyControllerFactory implements AbstractFactoryInterface
```
Ocramius: s/Factory/AbstractFactory

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function canCreate(ContainerInterface $container, $requestedName)
+    {
+        if (! is_string($requestedName) || ! class_exists($requestedName)) {
+            return false;
+        }
+
+        $implements = class_implements($requestedName);
```
Ocramius: No need for this assignment

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+
+        return $reflectionClass->newInstanceArgs($parameters);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function canCreate(ContainerInterface $container, $requestedName)
+    {
+        if (! is_string($requestedName) || ! class_exists($requestedName)) {
```
Ocramius: can `$requestedName` ever be a non-string?

src/Controller/LazyControllerFactory.php
snapshotpl: `::class` need to install all repositories, because need this classes

Ocramius: No, `::class` does not cause autoloading.

weierophinney: @snapshotpl Not true; PHP will resolve the class name using `::class` even if it cannot autoload the class.

Ocramius: You can use `::class` even with non-existing classes

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+class LazyControllerFactory implements AbstractFactoryInterface
+{
+    /**
+     * Maps known classes/interfaces to the service that provides them; only
+     * required for those services with no entry based on the class/interface
+     * name.
+     *
+     * @var string[]
+     */
+    private $aliases = [
```
snapshotpl: It will be nice to provide custom aliases

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,154 @@
...
+                    $requestedName,
+                    $parameter->getName(),
+                    $type
+                ));
+            }
+
+            $parameters[] = $container->get($type);
+        }
+
+        return $reflectionClass->newInstanceArgs($parameters);
```
snapshotpl: Why not use http://php.net/manual/en/migration56.new-features.php#migration56.new-features.splat if we support php `^5.6` ?

src/Controller/LazyControllerFactory.php
weierophinney: I've seen `null` a few times, though primarily when used in a servicemanager v2 context. I'll remove that check.

src/Controller/LazyControllerFactory.php
snapshotpl: Interesting. So `Zend\Console\Adapter\AdapterInterface::class` will be `Zend\Console\Adapter\AdapterInterface` string even I autoload it or not?

src/Controller/LazyControllerFactory.php
weierophinney: To clarify: zend-mvc v3 _requires_ zend-servicemanager v3, and the class is implementing only the v3 interface. Thus, the check is superfluous.

src/Controller/LazyControllerFactory.php
weierophinney: I'll make it protected so this factory can be extended, then.

src/Controller/LazyControllerFactory.php
```diff
@@ -0,0 +1,164 @@
...
+                continue;
+            }
+
+            if (! $parameter->getClass()) {
+                $parameters[] = null;
+                continue;
+            }
+
+            $type = $parameter->getClass()->getName();
+            $type = array_key_exists($type, $this->aliases) ? $this->aliases[$type] : $type;
```
snapshotpl: `isset` is faster, right?

src/Controller/LazyControllerAbstractFactory.php
```diff
@@ -0,0 +1,165 @@
...
+ * make assumptions about which argument(s) they might replace.
+ */
+class LazyControllerAbstractFactory implements AbstractFactoryInterface
+{
+    /**
+     * Maps known classes/interfaces to the service that provides them; only
+     * required for those services with no entry based on the class/interface
+     * name.
+     *
+     * Extend the class if you wish to add to the list.
```
snapshotpl, line 81: Can't we get this list from configuration?

weierophinney, line 81: I'd rather not add a configuration point for it. This currently lists all "special case" service names, and we recommend that users utilize fully qualified interface and/or class names for their service names.

If a user really wants to add more, they should have to think twice before doing so.

weierophinney:
@snapshotpl Thanks for reminding me about the splat operator; still not used to being able to use PHP 5.6 features!

snapshotpl:
That's why we goes into 5.6!, right?

src/Controller/LazyControllerAbstractFactory.php
snapshotpl, line 81: No I see that https://github.com/zendframework/zend-mvc/pull/165/files#diff-898a3a94db9fcef93eeeb26a23397ecaR139 will resolve many cases. Anyway this feature it's for really lazy developers, and never will be cover all cases

src/Controller/LazyControllerAbstractFactory.php
```diff
@@ -0,0 +1,177 @@
...
+            return new $requestedName();
+        }
+
+        $reflectionParameters = $constructor->getParameters();
+
+        if (empty($reflectionParameters)) {
+            return new $requestedName();
+        }
+
+        $parameters = array_map(function (ReflectionParameter $parameter) use ($container, $requestedName) {
```
snapshotpl: I read about performance `array_map` and `foreach` and looks a little bit slower. `use` also decrease readability for me. Maybe `foreach` will be better?

weierophinney: Micro-optimizations with the size of the arrays we have here. But I agree on the readability part; really wish you could do callables using private methods if invoked within the context of the class.

weierophinney: I've refactored this now. `resolveParameter()` now returns a closure over the container and requested name, leaving:

``` php
$parameters = array_map(
    $this->resolveParameter($container, $requestedName),
    $reflectionParameters
);
```

src/Controller/LazyControllerAbstractFactory.php
weierophinney, line 81: Precisely. If you need something more robust, or with more features, you should likely create a custom factory at that point, not extend this one. :smile:

Ocramius:
@weierophinney I just realized that this PR has nothing to do with laziness, but is instead centered on "automatic di". Should I open an issue about it, or can you still work on it?

weierophinney:
@ocramius — What's your idea? A rename? Or a more generic offering in
zend--servicemanager?

Ocramius:
A rename is sufficient. Just want to avoid messy situations with
documentation later on

Issue #182, opened 2016-07-05
[Test] InitializationIntegrationTest must shut up


sasezaki:
currently,  when running `phpunit`

https://travis-ci.org/zendframework/zend-mvc/jobs/141448300#L393

``` sh
......Application\Controller\PathController::dispatch.....................S...................................  63 / 368 ( 17%)
```

Should be shut up to avoid confusion.

This `Application\Controller\PathController::dispatch` echo is caused by Application/InitializationIntegrationTest.php

sasezaki:
Could be stop emit response 

``` php
        if ($application->getServiceManager()->has('SendResponseListener')) {
            /** @var SendResponseListener $sendResponseListener */
            $sendResponseListener = $application->getServiceManager()->get('SendResponseListener');
            $sendResponseListener->getEventManager()->clearListeners(SendResponseEvent::EVENT_SEND_RESPONSE);
        }
```

or 

``` php
        ob_start();
        $application->run();
        $content = ob_get_clean();
```

Test case method named as `testDefaultInitializationWorkflow`.
So, I will make PR with output buffer flush approach.

sasezaki:
#183 is merged. so I close this.

PR #183, opened 2016-07-05
#182 buffering application echo under testcase

sasezaki:
Test  code should be shut up to avoid confusion.
#182


weierophinney:
Thanks, @sasezaki - great to get rid of that output finally!

PR #184, opened 2016-07-11
Move unnecessary processBodyContent()

davidwindell:
This is wasting resource for delete's where this doesn't need to be processed


weierophinney:
Thanks, @davidwindell!

trburgess:
@davidwindell @weierophinney 

I know this is an old PR, I am just migrating our app to PHP7 and currently using ZF2 2.5.3. This breaks the whole app as when deleting a single resource there is never going to be a body present.

Is it possible to have this back ported to 2.5.*? This shouldn't break anything as `$data` is not passed as an argument to `$this->delete()` and therefore is not used at all.

PR #187, opened 2016-07-16
Remove typehint from marshalControllerNotFoundEvent

mjvanmiddelaar:
- Removes the tyephint for the `$exception` argument, to allow either
  `Exception` or `Throwable` to be provided.


weierophinney:
@mahkali Thanks for the patch! In the future, please issue bugfixes against the master branch, so we do not need to cherry-pick. Thanks!

PR #188, opened 2016-07-20
Application::run() return completeRequest()

samsonasik:
as completeRequest() already return `$this`


weierophinney:
Thanks, @samsonasik.

Issue #194, opened 2016-08-16
AcceptableViewModelSelector returns wrong View Model


MatthiasKuehneEllerhold:
I tried using the AcceptableViewModelSelector with this Accept-Header in my request (Its the standard header for HTML in FF): 
`text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`
## Expectation:

It should return "ViewModel" with the given Accept-Header and it should return a JsonModel with "application/json" in my Accept-Header.
### Testcase 1:

``` php
$viewModel = $this->acceptableViewModelSelector(
    [
        JsonModel::class => ['application/json'],
        FeedModel::class => ['application/rss+xml'],
    ]
);
```

**Returns**
- JsonModel if application/json
- FeedModel if application/rss+xml 
- JsonModel with the Accept-Header above
### Testcase 2:

``` php
$viewModel = $this->acceptableViewModelSelector(
    [
        FeedModel::class => ['application/rss+xml'],
        JsonModel::class => ['application/json'],
    ]
);
```

**Returns**
- JsonModel if application/json
- FeedModel if application/rss+xml 
- FeedModel with the Accept-Header above
### Testcase 3:

``` php
$viewModel = $this->acceptableViewModelSelector(
    [
        FeedModel::class => ['application/rss+xml'],
        JsonModel::class => ['application/json'],
        ViewModel::class => ['text/html'],
    ]
);
```

**Returns**
- JsonModel if application/json
- FeedModel if application/rss+xml 
- ViewModel with the Accept-Header above
## Observation

It seems like the fallback to defaultViewModel does not work and it'll returns the first Model it gets if no match was successfull.

MatthiasKuehneEllerhold:
PS: 
I tried it with
- PHP 7.0.9-1~dotdeb+8.1
- zendframework/zend-mvc 3.0.2
- zendframework/zend-view 2.8.1
- zendframework/zend-http 2.5.5

Possibly related to #16 ?

adamlundrigan:
I added this test to the zend-mvc test suite:

``` php
    public function testAppropriateFallbackOrderTestCase1()
    {
        $arr = [
            'Zend\View\Model\JsonModel' => ['application/json'],
            'Zend\View\Model\FeedModel' => ['application/rss+xml'],
        ];

        $header   = Accept::fromString('text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8');
        $this->request->getHeaders()->addHeader($header);
[...]

PR #195, opened 2016-08-16
Clarify AcceptableViewModelSelector parsing and browser behavior

MatthiasKuehneEllerhold:
See https://github.com/zendframework/zend-mvc/issues/194 for discussion.

### DIFF
--- a/doc/book/plugins.md
+++ b/doc/book/plugins.md
@@ -51,7 +51,7 @@ class SomeController extends AbstractActionController
         ],
         \Zend\View\Model\JsonModel::class => [
             'application/json',
-            'application/javascript'
+            'application/javascript',
         ],
         \Zend\View\Model\FeedModel::class => [
             'application/rss+xml',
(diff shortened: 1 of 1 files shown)


doc/book/plugins.md
```diff
@@ -40,35 +40,44 @@ As an example:
...
-   {
-      $viewModel = $this->acceptableViewModelSelector($this->acceptCriteria);
-
-      // Potentially vary execution based on model returned
-      if ($viewModel instanceof JsonModel) {
-         // ...
-      }
-   }
+    protected $acceptCriteria = [
+        // Make sure ViewModel is the first as a fallback
```
adamlundrigan: Every browser will send `text/html` at a higher priority than `*/*` so as long as a ViewModel with a HTML mimetype is in `$acceptCriteria` it will work.

MatthiasKuehneEllerhold: I've added this line to make sure that ViewModel will be the intended fallback for all unmatched parts.

adamlundrigan: Fair point.  I would then just change the wording slightly as it assumes that I want HTML as the fallback.  Maybe "The first model in the list will be used as the default in case of no exact match", or similar ?

MatthiasKuehneEllerhold:
Any news on this PR?

weierophinney:
Thank you, @MatthiasKuehneEllerhold.

PR #198, opened 2016-08-29
Use a factory to create the SendResponseListener

weierophinney:
Reported in:
- zendframework/zend-mvc-console#10
- zendframework/zend-mvc-console#11
- zendframework/zend-mvc-console#12

The `SendResponseListener` was lazy-instantiating an event manager on first request to `getEventManager()`. However, because initializers run after delegators, this meant that the EM instance composed did not have a shared EM instance, which triggered the initializer to re-inject, losing any listeners injected by delegators.

This patch introduces a factory for the `SendResponseListener`. The factory creates the instance, and then injects it with an EM instance pulled from the container; since these are guaranteed to have a shared EM instance, the initializer will skip injection, keeping any listeners injected by delegators.

### DIFF
--- a/src/Service/SendResponseListenerFactory.php
+++ b/src/Service/SendResponseListenerFactory.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * @link      http://github.com/zendframework/zend-mvc for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mvc\Service;
+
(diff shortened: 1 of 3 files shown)


GeeH:
Looks fine to me. 👍

weierophinney:
.03% is statistically insignificant. i really wish we could tell coveralls to not report failure when the threshold is that low. :angry:

weierophinney:
Thanks, @Geeh - I'll merge and tag with 3.0.3, now that I've merged the other PRs.

svycka:
@weierophinney but you can change this in coveralls settings if you want. There is two options:
- COVERAGE THRESHOLD FOR FAILURE
- COVERAGE DECREASE THRESHOLD FOR FAILURE
  you can set these as you wish.

PR #200, opened 2016-09-13
Move Coverage reporting to latest env

Koopzington:
Requiring php-coveralls results in a full update of all dependencies instead of composer using the versions mentioned in the composer.lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,12 +30,12 @@ matrix:
     - php: 5.6
       env:
         - DEPS=locked
-        - TEST_COVERAGE=true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,12 +30,12 @@ matrix:
     - php: 5.6
       env:
         - DEPS=locked
-        - TEST_COVERAGE=true
         - DEPLOY_DOCS="$(if [[ $TRAVIS_BRANCH == 'master' && $TRAVIS_PULL_REQUEST == 'false' ]]; then echo -n 'true' ; else echo -n 'false' ; fi)"
         - PATH="$HOME/.local/bin:$PATH"
     - php: 5.6
       env:
         - DEPS=latest
+        - TEST_COVERAGE=true
```
asgrim, line 38: Using `7.1` might be faster to report back coverage now? It looks like `7.1` isn't even tested in `.travis.yml` yet, I'll create a PR to add that separately.

PR #210, opened 2016-12-12
Add route match params to request attributes

asgrim:
Proposed change copies the `RouteMatch` parameters into the PSR-7 `\Zend\Diactoros\ServerRequest` so that they may be referred to in a middleware that has been invoked by the `MiddlewareListener`. Currently, if using a middleware in this way, I don't see any other way to access the matched route information (specifically, matched route parameters from a segment route, e.g. `/:uuid` matches, but I can't access the `uuid` parameter anywhere in the middleware action).

### DIFF
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -14,6 +14,7 @@ use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface;
 use Zend\Psr7Bridge\Psr7ServerRequest as Psr7Request;
 use Zend\Psr7Bridge\Psr7Response;
+use Zend\Router\RouteMatch;
 
 class MiddlewareListener extends AbstractListenerAggregate
 {
@@ -59,7 +60,7 @@ class MiddlewareListener extends AbstractListenerAggregate
 
         $caughtException = null;
         try {
-            $psr7Request = Psr7Request::fromZend($request);
+            $psr7Request = Psr7Request::fromZend($request)->withAttribute(RouteMatch::class, $routeMatch);
             foreach ($routeMatch->getParams() as $key => $value) {
                 $psr7Request = $psr7Request->withAttribute($key, $value);
             }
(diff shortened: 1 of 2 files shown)


akrabat:
Looks good to me.

src/MiddlewareListener.php
```diff
@@ -59,7 +59,11 @@ public function onDispatch(MvcEvent $event)
 
         $caughtException = null;
         try {
-            $return = $middleware(Psr7Request::fromZend($request), Psr7Response::fromZend($response));
+            $psr7Request = Psr7Request::fromZend($request);
+            foreach ($routeMatch->getParams() as $key => $value) {
```
michalbundyra, line 64: I think will be nice to inject also while `RouteMatch` (as we have it in expressive: https://github.com/zendframework/zend-expressive/blob/master/src/Application.php#L431-L435), not just individual matched params.

test/MiddlewareListenerTest.php
```diff
@@ -82,6 +88,28 @@ public function testSuccessfullyDispatchesMiddleware()
...
+        );
+
+        $this->routeMatch->getParams()->willReturn([
+            'myParam' => $matchedRouteParam,
+        ]);
+
+        $listener = new MiddlewareListener();
+        $return   = $listener->onDispatch($event);
+        self::assertInstanceOf(Response::class, $return);
+        self::assertSame($matchedRouteParam, $return->getBody());
```
michalbundyra: Please use `$this->` (instead of `self::`) in 2 above assertions to keep consistency with other tests.

michalbundyra (commented):
I think would be also nice inject whole RouteMatch, not only individual matched params (as we have it in expressive).

src/MiddlewareListener.php
asgrim: Thanks, added

test/MiddlewareListenerTest.php
asgrim: Updated

weierophinney (approved):
Patch looks quite reasonable, and makes the support within middleware more on par with that in MVC controllers. :+1:

PR #211, opened 2016-12-22
Remove support for servicemanager and eventmanager v2

snapshotpl:
zend-servicemanager v2 support is dropped in v3 so we can remove tone of code...

### DIFF
--- a/src/Service/ServiceListenerFactory.php
+++ b/src/Service/ServiceListenerFactory.php
@@ -10,14 +10,11 @@
 namespace Zend\Mvc\Service;
 
 use Interop\Container\ContainerInterface;
-use ReflectionClass;
-use Zend\Config\Config;
 use Zend\ModuleManager\Listener\ServiceListener;
 use Zend\ModuleManager\Listener\ServiceListenerInterface;
-use Zend\Mvc\Application;
 use Zend\Mvc\View;
 use Zend\ServiceManager\Exception\ServiceNotCreatedException;
-use Zend\ServiceManager\FactoryInterface;
+use Zend\ServiceManager\Factory\FactoryInterface;
 use Zend\ServiceManager\Factory\InvokableFactory;
 use Zend\ServiceManager\ServiceLocatorInterface;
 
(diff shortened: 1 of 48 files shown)


snapshotpl:
... and also zend-eventmanager v2.

src/Service/ServiceListenerFactory.php
```diff
@@ -96,20 +97,6 @@ class ServiceListenerFactory implements FactoryInterface
...
-     * Constructor
-     *
-     * When executed under zend-servicemanager v3, injects additional aliases
-     * to ensure backwards compatibility.
-     */
-    public function __construct()
-    {
-        $r = new ReflectionClass(ServiceLocatorInterface::class);
-        if ($r->hasMethod('build')) {
-            $this->injectV3Aliases();
```
Xerkus, line 108: breaking change, high chance someone still uses those compatibility aliases

src/Service/ServiceListenerFactory.php
```diff
@@ -263,22 +239,4 @@ private function validatePluginManagerOptions($options, $name)
...
-     * Inject additional aliases for zend-servicemanager v3 usage
-     *
-     * If the constructor detects that we're operating under zend-servicemanager v3,
-     * this method injects additional aliases to ensure that common services
-     * can be retrieved using both Titlecase and lowercase, and will get the
-     * same instances.
-     *
-     * @return void
-     */
-    private function injectV3Aliases()
```
Xerkus, line 277: breaking change, high chance someone still uses those compatibility aliases

Xerkus (changes requested):
Great stuff, however there are couple BC concerns.

src/Service/ServiceListenerFactory.php
snapshotpl: @Xerkus you have right. That's why I add aliases into `$defaultServiceConfig` property (above)

Xerkus (approved):

src/Service/ServiceListenerFactory.php
Xerkus: Indeed

Issue #196, opened 2016-08-21
[Question] Can we enable piped style middleware router for ZF MVC just like in Expressive?


gabbydgab:
I've tried mapping routes to middleware just like in the [docs](https://docs.zendframework.com/zend-mvc/middleware/#mapping-routes-to-middleware) but as of the moment it can only invoke one middleware action.

Is there a way we can integrate the [piped style middleware router](https://docs.zendframework.com/zend-expressive/cookbook/route-specific-pipeline/#what-about-pipeline-middleware-configuration) like in Zend Expressive? Or this feature is for the Router?

Ocramius:
Can probably produce a middleware pipe from the factory responsible of dispatching a certain middleware. Needs some fiddling, but you'd implement a middleware that wraps a set of middlewares, basically.

gabbydgab:
how sad T_T

still hoping for this to happen soon :)

PR #217, opened 2017-02-19
Modify middleware listener to use stratigility pipe

asgrim:
Taking issue #196 into play, this PR proposes to update the `MiddlewareListener` to use `Stratigility` to allow piped middleware. This means the `middleware` param can become an `array` and everything is pushed into the `MiddlewarePipe`... as let's be fair, only allowing just a single middleware is only marginally useful.

For this to work, `zendframework/zend-stratigility` needs to become a requirement for those using `MiddlewareListener` (I added it as `require-dev` for now), so could be seen as a BC break potentially; so happy to negotiate on how we'd like this added to `composer.json`. Otherwise, I believe (pending CI results) all the other tests still pass though.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,10 +24,11 @@
         "zendframework/zend-view": "^2.6.7"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.0.7 || ^5.7.14",
-        "zendframework/zend-coding-standard": "~1.0.0",
         "zendframework/zend-json": "^2.6.1 || ^3.0",
-        "zendframework/zend-psr7bridge": "^0.2"
+        "zendframework/zend-psr7bridge": "^0.2",
+        "fabpot/php-cs-fixer": "1.7.*",
+        "phpunit/phpunit": "^4.5",
+        "zendframework/zend-stratigility": "^2.0.1"
     },
     "suggest": {
         "zendframework/zend-json": "(^2.6.1 || ^3.0) To auto-deserialize JSON body content in AbstractRestfulController extensions, when json_decode is unavailable",
@@ -40,7 +41,8 @@
         "zendframework/zend-mvc-plugin-prg": "To provide Post/Redirect/Get functionality within controllers",
         "zendframework/zend-paginator": "^2.7 To provide pagination functionality via PaginatorPluginManager",
         "zendframework/zend-psr7bridge": "(^0.2) To consume PSR-7 middleware within the MVC workflow",
-        "zendframework/zend-servicemanager-di": "zend-servicemanager-di provides utilities for integrating zend-di and zend-servicemanager in your zend-mvc application"
+        "zendframework/zend-servicemanager-di": "zend-servicemanager-di provides utilities for integrating zend-di and zend-servicemanager in your zend-mvc application",
+        "zendframework/zend-stratigility": "zend-stratigility is required to use middleware pipes in the MiddlewareListener"
     },
     "config": {
         "sort-packages": true
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
     },
     "suggest": {
         "zendframework/zend-json": "(^2.6.1 || ^3.0) To auto-deserialize JSON body content in AbstractRestfulController extensions, when json_decode is unavailable",
+        "zendframework/zend-log": "^2.9.1  To provide log functionality via LogFilterManager, LogFormatterManager, and LogProcessorManager",
         "zendframework/zend-mvc-console": "zend-mvc-console provides the ability to expose zend-mvc as a console application",
         "zendframework/zend-mvc-i18n": "zend-mvc-i18n provides integration with zend-i18n, including a translation bridge and translatable route segments",
         "zendframework/zend-mvc-plugin-fileprg": "To provide Post/Redirect/Get functionality around forms that container file uploads",
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "phpunit/phpunit": "^6.0.7 || ^5.7.14",
         "zendframework/zend-coding-standard": "~1.0.0",
         "zendframework/zend-json": "^2.6.1 || ^3.0",
-        "zendframework/zend-psr7bridge": "^0.2"
+        "zendframework/zend-psr7bridge": "^0.2",
+        "zendframework/zend-stratigility": "^2.0.1"
     },
     "suggest": {
         "zendframework/zend-json": "(^2.6.1 || ^3.0) To auto-deserialize JSON body content in AbstractRestfulController extensions, when json_decode is unavailable",
@@ -40,7 +41,8 @@
         "zendframework/zend-mvc-plugin-prg": "To provide Post/Redirect/Get functionality within controllers",
         "zendframework/zend-paginator": "^2.7 To provide pagination functionality via PaginatorPluginManager",
         "zendframework/zend-psr7bridge": "(^0.2) To consume PSR-7 middleware within the MVC workflow",
-        "zendframework/zend-servicemanager-di": "zend-servicemanager-di provides utilities for integrating zend-di and zend-servicemanager in your zend-mvc application"
+        "zendframework/zend-servicemanager-di": "zend-servicemanager-di provides utilities for integrating zend-di and zend-servicemanager in your zend-mvc application",
+        "zendframework/zend-stratigility": "zend-stratigility is required to use middleware pipes in the MiddlewareListener"
     },
     "config": {
         "sort-packages": true
--- a/composer.json
+++ b/composer.json
@@ -24,10 +24,10 @@
         "zendframework/zend-view": "^2.6.7"
     },
     "require-dev": {
+        "phpunit/phpunit": "^6.0.7 || ^5.7.14",
+        "zendframework/zend-coding-standard": "~1.0.0",
         "zendframework/zend-json": "^2.6.1 || ^3.0",
         "zendframework/zend-psr7bridge": "^0.2",
-        "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/phpunit": "^4.5",
         "zendframework/zend-stratigility": "^2.0.1"
     },
     "suggest": {
--- a/src/Exception/MiddlewareNotCallableException.php
+++ b/src/Exception/MiddlewareNotCallableException.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mvc\Exception;
+
+final class MiddlewareNotCallableException extends RuntimeException
+{
+    /**
+     * @var string
+     */
+    private $middlewareName;
+
+    /**
+     * @param string $middlewareName
+     * @return self
+     */
+    public static function fromMiddlewareName($middlewareName)
+    {
+        $middlewareName = (string)$middlewareName;
+        $instance = new self(sprintf('Cannot dispatch middleware %s', $middlewareName));
+        $instance->middlewareName = $middlewareName;
+        return $instance;
+    }
+
+    /**
+     * @return string
+     */
+    public function toMiddlewareName()
+    {
+        return null !== $this->middlewareName ? $this->middlewareName : '';
+    }
+}
--- a/src/Exception/MiddlewareNotCallableException.php
+++ b/src/Exception/MiddlewareNotCallableException.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mvc\Exception;
+
+final class MiddlewareNotCallableException extends RuntimeException
+{
+    /**
+     * @var string
+     */
+    private $middlewareName;
+
+    /**
+     * @param string $middlewareName
+     * @return self
+     */
+    public static function fromMiddlewareName($middlewareName)
+    {
+        $middlewareName = (string)$middlewareName;
+        $instance = new self(sprintf('Cannot dispatch middleware %s', $middlewareName));
+        $instance->middlewareName = $middlewareName;
+        return $instance;
+    }
+
+    /**
+     * @return string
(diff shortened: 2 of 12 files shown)


asgrim:
Apparently I didn't have an up to date `develop` branch >.<

asgrim:
Rebased correctly now \o/

src/MiddlewareListener.php
```diff
@@ -64,7 +80,15 @@ public function onDispatch(MvcEvent $event)
             foreach ($routeMatch->getParams() as $key => $value) {
                 $psr7Request = $psr7Request->withAttribute($key, $value);
             }
-            $return = $middleware($psr7Request, Psr7Response::fromZend($response));
+            $return = $pipe(
+                $psr7Request,
+                $psr7ResponsePrototype,
+                function (PsrServerRequestInterface $request, PsrResponseInterface $response) {
+                    throw new ReachedFinalHandlerException(
```
Ocramius: @weierophinney are you OK with moving this to a static named constructor instead? Asking because consistency...

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
-            return $return;
+        $psr7ResponsePrototype = Psr7Response::fromZend($response);
+        $pipe = new MiddlewarePipe();
+        $pipe->setResponsePrototype($psr7ResponsePrototype);
+
+        $middlewaresToBePiped = !is_array($middleware) ? [$middleware] : $middleware;
+
+        $middlewareName = 'noMiddlewarePiped';
+        $middlewareToBePiped = null;
+        foreach ($middlewaresToBePiped as $middlewareToBePiped) {
```
Ocramius: Segregate to a private method.

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
-        }
-        if (! is_callable($middleware)) {
-            $return = $this->marshalMiddlewareNotCallable($application::ERROR_MIDDLEWARE_CANNOT_DISPATCH, $middlewareName, $event, $application);
-            $event->setResult($return);
-            return $return;
+        $psr7ResponsePrototype = Psr7Response::fromZend($response);
+        $pipe = new MiddlewarePipe();
+        $pipe->setResponsePrototype($psr7ResponsePrototype);
+
+        $middlewaresToBePiped = !is_array($middleware) ? [$middleware] : $middleware;
```
Ocramius: Don't negate: invert the ternary please

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
+
+        $middlewareName = 'noMiddlewarePiped';
+        $middlewareToBePiped = null;
+        foreach ($middlewaresToBePiped as $middlewareToBePiped) {
+            $middlewareName = is_string($middlewareToBePiped) ? $middlewareToBePiped : get_class($middlewareToBePiped);
```
Ocramius: Because [`get_class()` is stupid](https://3v4l.org/5O4QA) this check is insufficient

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
+
+            if (is_string($middlewareToBePiped) && $serviceManager->has($middlewareToBePiped)) {
```
Ocramius: `is_string()` check is duplicate

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
+                $middlewareToBePiped = $serviceManager->get($middlewareToBePiped);
+            }
+            if (! is_callable($middlewareToBePiped)) {
```
Ocramius: Should probably also **NOT** be a string?

Ocramius: An example is an incompatible function name being passed in.

src/MiddlewareListener.php
Ocramius: Probably should just return `$pipe` in the private method.

src/MiddlewareListener.php
```diff
@@ -75,7 +99,9 @@ public function onDispatch(MvcEvent $event)
             $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
             $event->setError($application::ERROR_EXCEPTION);
             $event->setController($middlewareName);
-            $event->setControllerClass(get_class($middleware));
+            if (null !== $middlewareToBePiped) {
+                $event->setControllerClass(get_class($middlewareToBePiped));
```
Ocramius: You always assume that the last piped middleware is the "responding" one here... Probably needs documenting/commenting

src/MiddlewareListener.php
```diff
@@ -47,15 +50,28 @@ public function onDispatch(MvcEvent $event)
...
-            $event->setResult($return);
-            return $return;
+        $psr7ResponsePrototype = Psr7Response::fromZend($response);
+        $pipe = new MiddlewarePipe();
+        $pipe->setResponsePrototype($psr7ResponsePrototype);
+
+        $middlewaresToBePiped = !is_array($middleware) ? [$middleware] : $middleware;
+
+        $middlewareName = 'noMiddlewarePiped';
+        $middlewareToBePiped = null;
```
Ocramius: Can an empty pipeline be considered valid?

Ocramius (dismissed):

src/MiddlewareListener.php
asgrim: I don't see why not; you'll just reach the final handler anyway...

src/MiddlewareListener.php
asgrim: Didn't change this from the original code...

src/MiddlewareListener.php
asgrim: If it's a string, it won't be a string any more. Also just following same logic that was already there...

src/MiddlewareListener.php
asgrim: Removed these as there's no longer just one middleware, so doesn't make sense. I guess this causes a definite BC break which is kinda rubbish though?

asgrim:
@Ocramius some changes applied as requested.

src/MiddlewareListener.php
asgrim: Not sure what to do about it; it's just the flexibility around using a string as the middleware (load from container), or a callable (not ideal, but allowed)

src/MiddlewareListener.php
asgrim: Done

src/MiddlewareListener.php
asgrim: Done

asgrim:
*shakes fist at PHP 5.6 support*

composer.json
```diff
@@ -27,7 +27,8 @@
         "zendframework/zend-json": "^2.6.1 || ^3.0",
         "zendframework/zend-psr7bridge": "^0.2",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/phpunit": "^4.5"
+        "phpunit/phpunit": "^4.5",
+        "zendframework/zend-stratigility": "^2.0"
```
weierophinney: This should likely be `^2.0.1`, but I can take care of that during merge.

src/MiddlewareListener.php
weierophinney: I am, but I'm not convinced it's needed, as there's a static message with no dynamic substitutions, and no additional behavior.

src/MiddlewareListener.php
```diff
@@ -74,8 +97,6 @@ public function onDispatch(MvcEvent $event)
         if ($caughtException !== null) {
             $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
             $event->setError($application::ERROR_EXCEPTION);
-            $event->setController($middlewareName);
-            $event->setControllerClass(get_class($middleware));
```
weierophinney, line 83: Why are these lines removed?

src/MiddlewareListener.php
```diff
@@ -96,6 +117,42 @@ public function onDispatch(MvcEvent $event)
...
+            if (null === $middlewareToBePiped) {
+                throw new \InvalidArgumentException('Middleware name cannot be null');
+            }
+
+            $middlewareName = is_string($middlewareToBePiped) ? $middlewareToBePiped : get_class($middlewareToBePiped);
+
+            if (is_string($middlewareToBePiped) && $serviceLocator->has($middlewareToBePiped)) {
+                $middlewareToBePiped = $serviceLocator->get($middlewareToBePiped);
+            }
+            if (! is_callable($middlewareToBePiped)) {
```
weierophinney: This needs to change to be:

```php
if (! $middlewareToBePiped instanceof \Http\Interop\ServerMiddleware\MiddlewareInterface
    && ! is_callable($middlewareToBePiped)
)
```

as zend-stratigility 2.0 allows either of those types. Testing only for `is_callable()` means that http-interop middleware cannot be used!

src/MiddlewareListener.php
weierophinney: One potential refactor I see here is to instead define `ReachedFinalHandlerException` as follows:

```php
class ReachedFinalHandlerException extends RuntimeException
{
    public static function createFinalHandler()
    {
        return new self('Reached the final handler for middleware pipe - check the pipe configuration');
    }

    public function __invoke()
    {
        throw $this;
    }
}
```

This would then allow you to do the following above:

```php
$return = $pipe($psr7Request, $psr7Response, ReachedFinalHandlerException::createFinalHandler());
```

Not sure if that's abusing inheritance and SRP, though. :grin:

src/MiddlewareListener.php
weierophinney: Once you make those changes, you'll also need to add tests for piping http-interop middleware.

weierophinney (changes requested):

weierophinney:
Also, @asgrim — You'll need to rebase against latest develop branch, as I pushed a bunch of QA tooling changes earlier (PHPUnit 5.7/6.0 support, in particular, as well as zend-coding-standards integration).

src/MiddlewareListener.php
asgrim: There is potentially no longer just "one" middleware name... how do we decide which is the failed one? First? Last? IIRC I don't think there's any way (outside of the exception bubble here)

weierophinney: Ah, right - that makes sense, then.

src/MiddlewareListener.php
asgrim: Yikes, I'm not really sure about the exception itself being the final handler, although I gotta say it's quite clever... does feel a bit wrong. I'll just refactor to named constructor for now and if you want to change it to be the actual final handler, I'm happy to :)

test/MiddlewareListenerTest.php
```diff
@@ -88,6 +92,38 @@ public function testSuccessfullyDispatchesMiddleware()
         $this->assertEquals('Test!', $return->getBody());
     }
 
+    public function testSuccessfullyDispatchesHttpInteropMiddleware()
+    {
+        $expectedOutput = uniqid('expectedOutput', true);
+
+        $event = $this->createMvcEvent('path', new class($expectedOutput) implements MiddlewareInterface {
```
asgrim: bah, doesn't work on PHP 5.6 - I thought `develop` branches were min PHP 7.1 now? Maybe I didn't read properly?

asgrim:
@weierophinney all updated, thanks for review!

src/MiddlewareListener.php
```diff
@@ -69,7 +80,13 @@ public function onDispatch(MvcEvent $event)
             foreach ($routeMatch->getParams() as $key => $value) {
                 $psr7Request = $psr7Request->withAttribute($key, $value);
             }
-            $return = $middleware($psr7Request, Psr7Response::fromZend($response));
+            $return = $pipe(
```
weierophinney: @asgrim So sorry I didn't notice this earlier, but... If we're pinning to Stratigility 2, we should likely use the http-interop `process()` method here instead of `__invoke()`, which means using a `DelegateInterface` instead of a callable.

May I suggest the following:

```php
$pipe->setResponsePrototype($psr7ResponsePrototype);
$return = $pipe->process($psr7Request, new \Zend\Stratigility\Delegate\CallableDelegateDecorator(
    function (PsrServerRequestInterface $request, PsrResponseInterface $response) {
        throw ReachedFinalHandlerException::create();
    }
));
```

That would accomplish the same as what you have here, and be forwards compatible.

src/Exception/MiddlewareNotCallableException.php
```diff
@@ -0,0 +1,38 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mvc\Exception;
+
+final class MiddlewareNotCallableException extends RuntimeException
```
weierophinney: Since we're allowing either callable or http-interop middleware, perhaps we should change this to `InvalidMiddlewareException` instead?

src/MiddlewareListener.php
```diff
@@ -101,6 +116,42 @@ public function onDispatch(MvcEvent $event)
...
+    private function createPipeFromSpec(
+        ContainerInterface $serviceLocator,
+        ResponseInterface $responsePrototype,
+        array $middlewaresToBePiped
+    ) {
+        $pipe = new MiddlewarePipe();
+        $pipe->setResponsePrototype($responsePrototype);
+        foreach ($middlewaresToBePiped as $middlewareToBePiped) {
+            if (null === $middlewareToBePiped) {
+                throw new \InvalidArgumentException('Middleware name cannot be null');
```
weierophinney: If we rename the exception to `InvalidMiddlewareException`, we could also add another named constructor: `::becauseNull()` (obviously, choose something better!).

weierophinney (changes requested):
Few last nit-picks, but looks great!

src/MiddlewareListener.php
asgrim: Yes, this seems reasonable, though I note that `__invoke` does the wrapping in the `CallableDelegateDecorator` for us, but I have no strong feelings either way. If `__invoke` is eventually going to go, then yes the switch makes sense!

`setResponsePrototype` is already called in `createPipeFromSpec` btw, so I don't think needs calling again (unless I missed something)

weierophinney: Ah, okay, nice.

As noted, using `process()` will be forwards compatible, meaning that when Stratigility adopts a finalized PSR-15, we likely can just add an `||` to the constraint and not need any actual code changes. I figure, do it once. :smile:

weierophinney (approved):
Per discussion in Slack, I'll perform the last requested changes now and merge.

Thanks, @asgrim!

weierophinney:
Thanks, @asgrim

PR #221, opened 2017-02-25
Enable PHP 7.1 testing on Travis CI

asgrim:
Wasn't sure to base this off `master` or `develop`, but happy to rebase either way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,6 @@ matrix:
     - php: 7.1
       env:
         - DEPS=locked
-        - CHECK_CS=true
     - php: 7.1
       env:
         - DEPS=latest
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -46,6 +46,16 @@ matrix:
     - php: 7
       env:
         - DEPS=latest
+    - php: 7.1
+      env:
+        - DEPS=lowest
+    - php: 7.1
+      env:
+        - DEPS=locked
+        - CHECK_CS=true
```
michalbundyra: Probably we don't need to run CS check on PHP7.1, usually we run it only on one version, and as it is on PHP7 here is redundant.

asgrim: :+1: removed

Xerkus:
@asgrim rebase to master

asgrim:
@Xerkus thanks, rebased to master

PR #222, opened 2017-02-27
phpunit 6 update

kokspflanze:
- phpunit 4 droped
- added phpunit 5.7.14 and phpunit 6.0.7
- added ::class syntax in tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,8 @@ install:
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS satooshi/php-coveralls:^1.0 ; fi
+  - if [[ $TRAVIS_PHP_VERSION =~ ^5.6 ]]; then travis_retry composer update $COMPOSER_ARGS --with-dependencies $LEGACY_DEPS ; fi
   - travis_retry composer install $COMPOSER_ARGS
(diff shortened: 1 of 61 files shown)


kokspflanze:
the php7 latest part should be fixed with https://github.com/zendframework/zend-eventmanager/pull/47

test/Application/BadControllerTrait.php
```diff
@@ -72,7 +73,7 @@ public function prepareApplication()
                     },
                 ],
                 'invokables' => [
-                    'Request'              => 'Zend\Http\PhpEnvironment\Request',
+                    'Request'              => Request::class,
```
froschdesign, line 76: The topic of this PR is "PHPUnit 6" and all these changes makes a review more difficult.

kokspflanze: if you want i can try to split it in 2 PRs

.travis.yml
```diff
@@ -74,6 +72,7 @@ install:
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS satooshi/php-coveralls:^1.0 ; fi
+  - if [[ $TRAVIS_PHP_VERSION =~ ^5.6 ]]; then travis_retry composer update $COMPOSER_ARGS --with-dependencies $LEGACY_DEPS ; fi
```
michalbundyra: Order here is important. This is not gonna work as expected if we will have composer.lock locked with PHPUnit 6.0.

I would suggest also extract `satooshi/php-coveralls` to variable, as we have it in other modules (please see for example some expressive modules).

kokspflanze: i can change it to https://github.com/zendframework/zend-expressive-fastroute/blob/master/.travis.yml if that is okay

michalbundyra: Yes, it is right. Please update. Thanks!

Issue #72, opened 2016-02-23
Failed test with zend-form 2.7


remicollet:
With zend-form 2.6, test suite is ok

With zend-form 2.7

There were 2 errors:

```
1) ZendTest\Mvc\Service\FormAnnotationBuilderFactoryTest::testCreateService
Missing argument 2 for Mock_FormElementManager_1b518e1e::injectFactory(), called in /builddir/build/BUILDROOT/php-zendframework-zend-mvc-2.6.2-1.fc20.remi.x86_64/usr/share/php/Zend/Mvc/Service/FormAnnotationBuilderFactory.php on line 32 and defined

/builddir/build/BUILDROOT/php-zendframework-zend-mvc-2.6.2-1.fc20.remi.x86_64/usr/share/php/Zend/Mvc/Service/FormAnnotationBuilderFactory.php:32
/builddir/build/BUILD/zend-mvc-9babf3f501f7db532f3293ec39593237f297a7d0/test/Service/FormAnnotationBuilderFactoryTest.php:28

2) ZendTest\Mvc\Service\FormAnnotationBuilderFactoryTest::testCreateServiceSetsPreserveDefinedOrder
Missing argument 2 for Mock_FormElementManager_1b518e1e::injectFactory(), called in /builddir/build/BUILDROOT/php-zendframework-zend-mvc-2.6.2-1.fc20.remi.x86_64/usr/share/php/Zend/Mvc/Service/FormAnnotationBuilderFactory.php on line 32 and defined

/builddir/build/BUILDROOT/php-zendframework-zend-mvc-2.6.2-1.fc20.remi.x86_64/usr/share/php/Zend/Mvc/Service/FormAnnotationBuilderFactory.php:32
/builddir/build/BUILD/zend-mvc-9babf3f501f7db532f3293ec39593237f297a7d0/test/Service/FormAnnotationBuilderFactoryTest.php:42
```

weierophinney:
Confirmed. The issue is that initializers are expected to receive two arguments, but that factory is only using one; the tests are incorrect. Essentially, the zend-form update exposed a problem in the zend-mvc test suite!

weierophinney:
[...]

PR #223, opened 2017-02-28
Migration to latest zendframework/zend-form

boesing:
There is an issue with the `zendframework/zend-form` ^2.8.2 since the `FormElementManagerV2Polyfill` has reverse order of `injectFactory` which does not work with `zendframework/zend-mvc` `FormAnnotationBuilderFactory`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "zendframework/zend-eventmanager": "^2.6.2 || ^3.0",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
         "zendframework/zend-hydrator": "^1.1 || ^2.1",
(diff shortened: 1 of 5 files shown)


boesing:
Fixes #224

boesing:
@weierophinney Would u mind this as hotfix like #72?

boesing:
Well, actually we have to wait for zendframework/zend-form#143 before merging this.
After merging zendframework/zend-form#143, we have to raise version of `zendframework/zend-form` in `composer.json`.

boesing:
@Xerkus could you have a look at this or probably ping someone who might be responsible? Would like to use those changes in production.

weierophinney:
Thanks, @boesing

PR #229, opened 2017-04-07
Remove Zend\Version no longer makes sense anymore

svycka:
because Zend\Version does not get any attention https://github.com/zendframework/zend-version/pull/15 https://github.com/zendframework/zend-version/issues/10 and more.. I will try here :D

this is for release-2.7 only


weierophinney:
Thanks, @svycka

PR #233, opened 2017-04-20
Ensure that we replace `zendframework/zend-router` due to class confl…

boesing:
…icts

`zendframework/zend-mvc` in version less than 3.0 is still replacing `zendframework/zend-router` due to incompatible exception changes. Therefore we should tell projects that this package is replacing `zendframework/zend-router`.
The downsides are the mentioned exception incompatibilities but even if a user is requiring `zendframework/zend-router` in version 2 and using this package as well, he wont get the expected exceptions since `zendframework/zend-mvc` is being used instead of `zendframework/zend-router`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,9 @@
         "phpunit/PHPUnit": "^4.5",
         "sebastian/version": "^1.0.4"
     },
+    "replace": {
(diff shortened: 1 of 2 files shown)


weierophinney:
Thanks, @boesing

PR #234, opened 2017-04-27
Ensure proper PHPDoc

boesing:
Since always(?), the `AbstractActionController::notFoundAction` and the `AbstractActionController::indexAction` return `ViewModel`/`ConsoleModel` instead of `array`.

~~Aswell, I've removed deprecated methods which are already replaced by controller plugins aswell.~~

### DIFF
--- a/src/Controller/AbstractActionController.php
+++ b/src/Controller/AbstractActionController.php
@@ -12,6 +12,7 @@ namespace Zend\Mvc\Controller;
 use Zend\Http\Response as HttpResponse;
 use Zend\Mvc\Exception;
 use Zend\Mvc\MvcEvent;
+use Zend\View\Model\ConsoleModel;
 use Zend\View\Model\ViewModel;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Controller/AbstractActionController.php
```diff
@@ -85,26 +87,4 @@ public function onDispatch(MvcEvent $e)
...
         return $actionResponse;
     }
-
-    /**
-     * @deprecated please use the {@see \Zend\Mvc\Controller\Plugin\CreateHttpNotFoundModel} plugin instead: this
-     *             method will be removed in release 2.5 or later.
-     *
-     * {@inheritDoc}
-     */
-    protected function createHttpNotFoundModel(HttpResponse $response)
```
weierophinney: While these methods _should_ have been removed for the 3.0 release, they were not, and, as such, doing so now is a backwards compatibility break. Please create a separate PR for these removals which we can tag for the 4.0 release (which will target PHP 7.1).

Thanks!

weierophinney (changes requested):

src/Controller/AbstractActionController.php
boesing: Actually they were removed in #99 
I will revert the commit on this for release-2.7 tho.

weierophinney:
Thanks, @boesing

PR #236, opened 2017-04-28
Fix: allow middleware dispatch to behave like controller dispatch

Ocramius:
This patch makes sure that the successful dispatch of a middleware won't trigger exception and 404 strategies. This can happen when the return value of the `MiddlewareListener` is not a `Zend\Stdlib\ResponseInterface` instance, and therefore the `EventManager` will not short-circuit and stop event propagation. 

~~~This allows returning anything from a configured `middleware`:~~~

 * ~~~`['foo' => 'bar']`~~~
 * ~~~`'Hello World!'`~~~
 * ~~~`new ViewModel(['foo' => 'bar'])`~~~

~~~This also fixes https://github.com/zendframework/zendframework/issues/5806~~ ERRATA: it doesn't, as #217 now enforces middleware to return `ResponseInterface` instances.

~~~I'm still stuck on this though. As per tradition `Zend\Mvc\ViewManager` is being absolutely useless, and [attaches all listeners to the shared manager `DispatchableInterface` identifier](https://github.com/zendframework/zend-mvc/blob/2c64d2ed50063e13b23f1c7e19f17ce3354bea36/src/View/Http/ViewManager.php#L124-L152): that means that basically nothing in the view layer works when using this stuff.~~~

~~~Currently preparing a workaround that triggers events that have the `DispatchableInterface` identifier, since otherwise it is impossible to get this thing running correctly.~~~

### DIFF
--- a/src/Controller/MiddlewareController.php
+++ b/src/Controller/MiddlewareController.php
@@ -57,32 +57,14 @@ final class MiddlewareController extends AbstractController
 
     /**
      * {@inheritDoc}
-     * @throws \Zend\Mvc\Exception\RuntimeException
+     *
+     * @throws RuntimeException
      */
     public function onDispatch(MvcEvent $e)
     {
-        $request  = $this->request;
-        $response = $this->response;
-
-        if (! $request instanceof Request) {
-            throw new RuntimeException(sprintf(
-                'Expected request to be a %s, %s given',
-                Request::class,
-                get_class($request)
-            ));
-        }
-
-        if (! $response instanceof Response) {
-            throw new RuntimeException(sprintf(
-                'Expected response to be a %s, %s given',
-                Response::class,
-                get_class($response)
-            ));
-        }
-
         $routeMatch  = $e->getRouteMatch();
         $psr7Request = $this->populateRequestParametersFromRoute(
-            Psr7ServerRequest::fromZend($request)->withAttribute(RouteMatch::class, $routeMatch),
+            $this->loadRequest()->withAttribute(RouteMatch::class, $routeMatch),
             $routeMatch
         );
 
@@ -96,6 +78,26 @@ final class MiddlewareController extends AbstractController
         $e->setResult($result);
 
         return $result;
+    }
+
+    /**
+     * @return \Zend\Diactoros\ServerRequest
+     *
+     * @throws RuntimeException
+     */
+    private function loadRequest()
+    {
+        $request = $this->request;
+
+        if (! $request instanceof Request) {
+            throw new RuntimeException(sprintf(
+                'Expected request to be a %s, %s given',
+                Request::class,
+                get_class($request)
+            ));
+        }
+
+        return Psr7ServerRequest::fromZend($request);
     }
 
     /**
--- a/src/Controller/MiddlewareController.php
+++ b/src/Controller/MiddlewareController.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Mvc\Controller;
+
+use Zend\EventManager\EventManager;
+use Zend\Http\Request;
+use Zend\Http\Response;
+use Zend\Mvc\Exception\RuntimeException;
+use Zend\Mvc\MvcEvent;
+use Zend\Psr7Bridge\Psr7Response;
+use Zend\Psr7Bridge\Psr7ServerRequest;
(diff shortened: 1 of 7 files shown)


src/MiddlewareListener.php
```diff
@@ -91,6 +90,8 @@ public function onDispatch(MvcEvent $event)
             }
         }
 
+        $event->setError('');
```
Ocramius, line 110: Note: this still means that the `DispatchListener` will try to execute and fail. The patch is not yet done. Ideally, the `DispatchListener` should check whether there is a `$event->getResult()`, and bail out if there already is one.

weierophinney: I'd argue both the `DispatchListener` and `MiddlewareListener` should act that way, with the following at the top of the listeners:

```php
if (null !== $e->getResult()) {
    return;
}
```

Do you want to incorporate that change in this PR?

src/Controller/MiddlewareController.php
```diff
@@ -0,0 +1,83 @@
...
+ * @internal hack to make middleware trigger 'dispatch' events attached to the DispatchableInterface identifier.
+ */
+final class MiddlewareController extends AbstractController
+{
+    /**
+     * @var callable
+     */
+    private $middleware;
+
+    public function __construct(callable $middleware, EventManager $eventManager, MvcEvent $event)
```
weierophinney: Just a note: on the develop branch, we're now targeting Stratigility `^2.0.1`, which means we can have:

- callable middleware
- http-interop middleware

If this patch targets master, we would need to remove the typehint for `$middleware`, and also update how middleware is invoked based on whether or not it implements `Interop\Http\ServerMiddleware\MiddlewareInterface`.

src/Controller/MiddlewareController.php
```diff
@@ -0,0 +1,83 @@
...
+        $routeMatch  = $e->getRouteMatch();
+        $psr7Request = Psr7ServerRequest::fromZend($request)->withAttribute(RouteMatch::class, $routeMatch);
+
+        if ($routeMatch) {
+            foreach ($routeMatch->getParams() as $key => $value) {
+                $psr7Request = $psr7Request->withAttribute($key, $value);
+            }
+        }
+
+        $result = \call_user_func($this->middleware, $psr7Request, Psr7Response::fromZend($response));
```
weierophinney: See previous note about http-interop.

test/MiddlewareListenerTest.php
```diff
@@ -76,6 +84,7 @@ public function testSuccessfullyDispatchesMiddleware()
         $application = $event->getApplication();
 
         $application->getEventManager()->attach(MvcEvent::EVENT_DISPATCH_ERROR, function ($e) {
+            die(var_dump($e->getParam('exception')->getMessage()));
```
weierophinney: Um, `die()`? Why not just a `var_dump()` or capturing that info to a variable to emit with the `fail()` below?

weierophinney (changes requested):
I like this approach a ton; I think triggering events for dispatched middleware in the same way as for controllers could have a ton of benefits.

A few comments:

- This is technically new functionality (new class), and I think it would be best to do for the 3.1.0 release. As such, you'd need to rebase against develop... which has a number of changes due to me merging #217 earlier today. I'd be happy to help with that if you wish.
- The `MiddlewareController` should also allow http-interop middleware; I have detailed comments on that below.
- I agree that this and the `DispatchListener` should return early if the `MvcEvent` composes a result already, and think that change should be part of this PR (as this PR is the one that demonstrates the need for that change).

src/MiddlewareListener.php
Ocramius: @weierophinney done for both listeners

test/MiddlewareListenerTest.php
Ocramius: This is also gone :-)

Ocramius:
> This is technically new functionality (new class), and I think it would be best to do for the 3.1.0 release. As such, you'd need to rebase against develop... which has a number of changes due to me merging #217 earlier today. I'd be happy to help with that if you wish.

Provided

 > The `MiddlewareController` should also allow http-interop middleware; I have detailed comments on that below.

That's also handled. The `MiddlewareController`, being an internal class, can be changed as much as it suits us.

 > I agree that this and the `DispatchListener` should return early if the `MvcEvent` composes a result already, and think that change should be part of this PR (as this PR is the one that demonstrates the need for that change).

Implemented/tested

src/Controller/MiddlewareController.php
```diff
@@ -0,0 +1,122 @@
...
+    private $pipe;
+
+    /**
+     * @var ResponseInterface
+     */
+    private $responsePrototype;
+
+    public function __construct(
+        MiddlewarePipe $pipe,
+        ResponseInterface $responsePrototype,
```
Ocramius, line 47: Not particularly proud of this, but PSR-7's immutability makes it quite acceptable to pass in a request as a prototype. Still, this class is designed as "fire once" and "internal" specifically due to this weirdness.

PR #245, opened 2017-06-13
Fixes #242 : apply zend-console as require in 2.7

### LINKED ISSUE #242: zend-console required in 2.7.12
I'm still running projects on zend-mvc 2.7.* and tried to remove zend-console from one of them.
After doing so i was presented an error telling me that Zend\Console\Console was not found.
[The RequestFactory apparenty needs zend-console](https://github.com/zendframework/zend-mvc/blob/release-2.7.12/src/Service/RequestFactory.php#L48) while in 3.0 everything console related got removed.
[Since zend-console is only in require-dev](https://github.com/zendframework/zend-mvc/blob/release-2.7/composer.json#L28) we should maybe consider moving it to 'require' for the 2.7 branch... or just do nothing at all and tell people to upgrade to 3.0 already ;)

samsonasik:
Fixes #242


samsonasik:
@froschdesign this should be in `2.7.*` milestone instead of `3.1.1` ?

samsonasik:
rebased. not sure about re-generating composer.lock, @weierophinney could you verify ?

weierophinney:
Thanks, @samsonasik. I regenerated the composer.lock during merge.

PR #247, opened 2017-06-14
Fixes #246 : Forward plugin should detach problem listeners

### LINKED ISSUE #246: [ZF3] Forward plugin doesn't detach problem listeners anymore.
Looks like there was a regressions as part of a [v2 compatibility removable commit](https://github.com/zendframework/zend-mvc/commit/64c766d40c604e8a51523c1bc638453cd21acc5b#diff-a4e7671f65fba054e5c02177656ec841L212). The detachProblemListeners code builds the list of listeners to remove but doesn't actually remove them. For me, this manifests as my resulting ViewModel being double-rendered due to the InjectViewModelListener not actually being detached.

I will submit a PR to restore the block of code that actually detaches the listener.

nathanjosiah:
Fixes #246

### DIFF
--- a/src/Controller/Plugin/Forward.php
+++ b/src/Controller/Plugin/Forward.php
@@ -194,6 +194,16 @@ class Forward extends AbstractPlugin
                         if (! is_object($currentCallback)) {
                             continue;
                         }
+
+                        foreach ($classArray as $class) {
+                            if ($currentCallback instanceof $class) {
+                                // Pass $currentEvent; when using zend-eventmanager v2,
+                                // this is the CallbackHandler, while in v3 it's
+                                // the actual listener.
+                                $this->detachSharedListener($id, $currentEvent, $sharedEvents);
+                                $results[$id][$eventName][$priority] = $currentEvent;
+                            }
+                        }
                     }
                 }
             }
--- a/src/Controller/Plugin/Forward.php
+++ b/src/Controller/Plugin/Forward.php
@@ -197,9 +197,6 @@ class Forward extends AbstractPlugin
 
                         foreach ($classArray as $class) {
                             if ($currentCallback instanceof $class) {
-                                // Pass $currentEvent; when using zend-eventmanager v2,
-                                // this is the CallbackHandler, while in v3 it's
-                                // the actual listener.
                                 $this->detachSharedListener($id, $currentEvent, $sharedEvents);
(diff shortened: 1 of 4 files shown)


src/Controller/Plugin/Forward.php
```diff
@@ -194,6 +194,16 @@ protected function detachProblemListeners(SharedEvents $sharedEvents)
                         if (! is_object($currentCallback)) {
                             continue;
                         }
+
+                        foreach ($classArray as $class) {
+                            if ($currentCallback instanceof $class) {
+                                // Pass $currentEvent; when using zend-eventmanager v2,
```
froschdesign: Please remove this comment, it was only relevant for v2.

test/Controller/Plugin/ForwardTest.php
```diff
@@ -237,6 +238,35 @@ function ($e) {}
         $this->assertEquals(['content' => 'ZendTest\Mvc\Controller\TestAsset\ForwardController::testAction'], $result);
     }
 
+    public function testProblemListenersAreDetachedAndReattachedWhenPluginDispatchsRequestedController()
```
froschdesign: No assertion or expected exception?

froschdesign (changes requested):

test/Controller/Plugin/ForwardTest.php
nathanjosiah: @froschdesign I'm not sure what the ZF policy or practice is for this type of assertion but in this case, the assertion is implied in the mock method invocation expectation:
```php
$sharedEvents->expects($this->once())->method('detach')
```

If the method isn't invoked with the expected arguments, the test will fail with an exception. What is the preferred way of testing something like this? Maybe a test spy that monitors called methods and then the assertion can be the method was called?

nathanjosiah: @froschdesign Just bumping this, it's been 3 months and I'd love to start using ZF3. This issue is the only thing holding us back. 

I'm happy to make whatever change to the test. As my comment above says, my assertion is implied in the mock `once` invocation expectation. Would you prefer I change this to an explicit test spy? I'm not sure what the ZF test expectation would be for this use case.

Xerkus:
Test failure is unrelated and fixed in https://github.com/zendframework/zend-mvc/pull/260

PR #250, opened 2017-07-18
replace strstr with strpos

samsonasik:
based on php manual at http://php.net/manual/en/function.strstr.php
```
 If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead. 
```


samsonasik:
it seems it was on purpose https://3v4l.org/5MX6d , closing

samsonasik:
re-open, use `false !==` for the replaced to `strpos` usage https://3v4l.org/gkHYg

froschdesign (approved):

PR #251, opened 2017-07-25
RESTful: Handle single field request with empty value

autowp:
`Controller/AbstractRestfulController`

There is no way to PUT/PATCH object with single field to empty value.

```
PUT /item/1
Content-Type: application/x-www-form-urlencoded

name=
```

Request payload is interpreted as string (`$data = 'name=';`) regardless Content-Type.
Proper value is `$data = ['name' => ''];`

I think ideally is split behaviour by Content-Type but that follows to huge BC. (see [testCanReceiveStringAsRequestContent](/zendframework/zend-mvc/blob/master/test/Controller/RestfulControllerTest.php#L103))

My pull request handle just one situation (example above) by header.

### DIFF
--- a/test/Controller/RestfulControllerTest.php
+++ b/test/Controller/RestfulControllerTest.php
@@ -561,4 +561,29 @@ class RestfulControllerTest extends TestCase
             ['PUT',     json_encode(['foo' => 1]),      []],          // AbstractRestfulController::replaceList()
         ];
     }
+
+    /**
+     * @dataProvider providerParseSingleFieldObjectWithEmptyValueProvider
+     */
+    public function testParseSingleFieldObjectWithEmptyValue($method)
+    {
+        $entity = ['name' => ''];
+        $string = http_build_query($entity);
+        $this->request->setMethod($method)
+            ->setContent($string);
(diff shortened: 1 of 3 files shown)


test/Controller/RestfulControllerTest.php
```diff
@@ -100,22 +100,6 @@ public function testDispatchInvokesCreateMethodWhenNoActionPresentAndPostInvoked
         $this->assertEquals('create', $this->routeMatch->getParam('action'));
     }
 
-    public function testCanReceiveStringAsRequestContent()
```
michalbundyra, line 103: This test has been merged into the new test added below.

michalbundyra:
@autowp I've made the change a bit more generic. It does not depend now on the content-type, as it was before. In case we have one element with empty value we are checking if the last character of the content is `=`. If so, then it is "correct" urlencoded content and we return array, as expected.
In case we have any string without `=` it will return this string, not an array.

In general I think it should be completely changed to work properly with the given content type correctly, as now this is quite hacky, imo.

weierophinney:
This repository has been closed and moved to laminas/laminas-mvc; a new issue has been opened at https://github.com/laminas/laminas-mvc/issues/18.

weierophinney:

PR #254, opened 2017-09-23
Bugfix strtolower template name

kaansef:
Hi,

When i install Zend Framework 3 on ubuntu or ubuntu based operating systems, an exception new pages, "cannot resolve Application/Index/index" template. I found this bug's fix. I think, this problem occured so UTF-8.

### DIFF
--- a/src/View/Http/InjectTemplateListener.php
+++ b/src/View/Http/InjectTemplateListener.php
@@ -157,7 +157,7 @@ class InjectTemplateListener extends AbstractListenerAggregate
         }
 
         $name = preg_replace($pattern, $replacement, $name);
-        return strtolower($name);
+        return mb_strtolower($name);
     }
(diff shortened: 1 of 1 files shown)


Ocramius:
@kaansef there are two problems with this patch:

 * [ ] missing tests
 * [ ] this patch does not have a direct dependency on `ext-mbstring`, while your code assumes that it is available

kaansef:
Hi again,

I'm Turkish, and my English very bad. So sorry. Sorry for my pull request's doesn't use format. I think this bug only some Linux distros (UTF-8 used). Thank for reply.

Ocramius:
@kaansef can you add a test to demonstrate the problem?

michalbundyra:
Closing due to inactivity. Missing test case.

PR #258, opened 2017-11-06
Detach InjectViewModelListener before forwarding.

demiankatz:
A long time ago, this commit was introduced to the Forward plugin: https://github.com/zendframework/zendframework/commit/0b8fc7adb6b138feec99eaff8e7a2ef4fe6aa041. This prevents a problem where forwarding causes multiple InjectViewModelListeners to get attached, which in turn causes the view template to get rendered multiple times. This is a waste of cycles, and in some cases causes malfunctions (in situations where rendering a template has side effects).

It appears that somewhere along the line, this functionality was broken and the test intended to cover it was removed. This PR restores the broken functionality and adds a test to help catch future regressions.

Real life use case: my application has a "flash messages" view helper which checks the session for pending messages, displays them, and then clears the queue from the session. This bug prevents flash messages from displaying whenever the application performs a forward action, since the first render clears the queue, then the second render overwrites the first with no flash message data. For me, this is a critical bug.

Please let me know if there are any questions or if I can help to further refine this code. Thank you!

### DIFF
--- a/src/Controller/Plugin/Forward.php
+++ b/src/Controller/Plugin/Forward.php
@@ -184,15 +184,11 @@ class Forward extends AbstractPlugin
                 foreach ($events as $priority => $currentPriorityEvents) {
                     foreach ($currentPriorityEvents as $currentEvent) {
                         $currentCallback = $currentEvent;
-
-                        // If we have an array, grab the object
-                        if (is_array($currentCallback)) {
-                            $currentCallback = array_shift($currentCallback);
-                        }
-
-                        // This routine is only valid for object callbacks
-                        if (! is_object($currentCallback)) {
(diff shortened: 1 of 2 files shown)


demiankatz:
Note that the failing Travis tests here do not appear to be directly related to my changes. (It looks like some middleware-related dependency may have released a new version that removes an interface and breaks some tests).

Xerkus:
I wonder who was that idiot who reviewed that change and allowed it to pass.

In my defence it was a big PR and there were no tests for detaching listeners when I merged it.

Xerkus:
Duplicate of #247

demiankatz:
Thanks for sorting it out! :-)

PR #260, opened 2017-11-17
Add webimpress/http-middleware-compatibility support

Xerkus:
Add webimpress/http-middleware-compatibility support to make zend-mvc fully compatible with zend-stratigility 2.1

### DIFF
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -12,11 +12,12 @@ use Interop\Http\ServerMiddleware\MiddlewareInterface;
 use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface as PsrServerRequestInterface;
+use Webimpress\HttpMiddlewareCompatibility\MiddlewareInterface as CompatMiddleware;
 use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface;
+use Zend\Mvc\Controller\MiddlewareController;
 use Zend\Mvc\Exception\InvalidMiddlewareException;
 use Zend\Mvc\Exception\ReachedFinalHandlerException;
-use Zend\Mvc\Controller\MiddlewareController;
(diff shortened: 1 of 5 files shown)


src/MiddlewareListener.php
```diff
@@ -142,7 +143,10 @@ private function createPipeFromSpec(
             if (is_string($middlewareToBePiped) && $serviceLocator->has($middlewareToBePiped)) {
                 $middlewareToBePiped = $serviceLocator->get($middlewareToBePiped);
             }
-            if (! $middlewareToBePiped instanceof MiddlewareInterface && ! is_callable($middlewareToBePiped)) {
+            if (! $middlewareToBePiped instanceof MiddlewareInterface
```
michalbundyra, line 146: We don't need this line anymore, please check line 11 in this file:
```php
use Interop\Http\ServerMiddleware\MiddlewareInterface;
```

it should be removed.

You are using MiddlewareInstance from `webimpress/http-middleware-compatibility` so it's enough.

Xerkus: It is needed for stratigility 2.0.x

michalbundyra: @Xerkus ok, fine, but we are not testing it then... Maybe we should set requirements to ^2.0.1 as it was and write additional test?

michalbundyra: @Xerkus I've added tests in #262

Xerkus: Not really _supporting_ it any more but allowing it to be used anyway as it is an optional dependency. If it was direct dependency, I would have bumped it to ^2.1.x and dropped that check.

michalbundyra: As we keep this condition here, it should be also tested, this is why I've added it in PR #262 :)

weierophinney:
I've written up a proposal for how we should go about updating to PSR-15 on the forums:

- https://discourse.zendframework.com/t/psr-15-compatibility-issues-and-proposal/378

Please read and comment there before moving forwards with the current pull request.

Xerkus:
Since PSR-15 is out I am going to close this as won't fix.

PR #261, opened 2017-11-17
Docs update + Travis CI fix

michalbundyra:
Docs updated using `zf-maintainer rebase-doc-templates` tool

See: https://github.com/zendframework/maintainers#maintainer-tools


.travis.yml
```diff
@@ -29,8 +29,7 @@ matrix:
     - php: 7
       env:
         - DEPS=locked
-        - CS_CHECK=true
-        - TEST_COVERAGE=true
+        - LEGACY_DEPS="phpunit/phpunit"
```
Xerkus, line 32: ugh, what? for php 7?

michalbundyra: Yes, because if you have lock file with PHPUnit 6.4 generated on PHP 7.1 you have also installed some dependencie which requires PHP 7.1 - for example `doctrine/instantiator: 1.1.0` requires PHP 7.1.
It's easier to add whole phpunit to legacy dependencies and update it on PHP 7.0 as well than look for all of these phpunit-related dependencies and list them here.

PR #262, opened 2017-11-17
Tests for #260 - Stratigility 2.0 and 2.1

michalbundyra:
@Xerkus Please have a look


michalbundyra:
As #260 is closed, this also is no longer required.

PR #272, opened 2018-01-30
Use callable type hint instead of a comment

jameskraai:
Use callable type hint instead of a comment

### DIFF
--- a/src/Controller/AbstractRestfulController.php
+++ b/src/Controller/AbstractRestfulController.php
@@ -527,7 +527,7 @@ abstract class AbstractRestfulController extends AbstractController
      * @param  Callable $handler
      * @return AbstractRestfulController
      */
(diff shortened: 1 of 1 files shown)


Xerkus:
callable typehint is not available in php 5.6 which is the minimum supported version
https://github.com/zendframework/zend-mvc/blob/0cb6770856885ca916b44798e3ae37547a20b46e/composer.json#L20

Ocramius:
`callable` is supported since PHP 5.4

Xerkus:
It is. I confused it for iterable. Deleted comment immediately but mail already went out.

Xerkus:
Adding a parameter typehint could be a BC break if someone extended and overrode the method. Merging it for the next major.
@jameskraai thank you

PR #278, opened 2018-04-27
Move unnecessary processBodyContent()

chaos0815:
this commit copies the changes from davidwindell in 3.1.1 to the 2.7 branch.

not only for the performance but also because of the php7 problem of json_decode on an empty string.


Xerkus:
2.7 is no longer supported. See https://framework.zend.com/long-term-support

PR #280, opened 2018-05-07
change substr() check with strrpos at InjectTemplateListener::deriveControllerClass()

samsonasik:
- [x] Is this related to quality assurance?

### DIFF
--- a/src/View/Http/InjectTemplateListener.php
+++ b/src/View/Http/InjectTemplateListener.php
@@ -174,7 +174,7 @@ class InjectTemplateListener extends AbstractListenerAggregate
 
         $length = strlen($controller);
         if ($length > 10
-            && strrpos($controller, 'Controller') === ($length - 10)
+            && strrpos($controller, 'Controller') === $length - 10
         ) {
             $controller = substr($controller, 0, -10);
         }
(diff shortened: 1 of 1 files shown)


samsonasik:
after some check, it is actually needed, ref https://3v4l.org/umaZC

samsonasik:
strrpos still possible, but with keep the strlen() usage, ref https://3v4l.org/I2N7t

src/View/Http/InjectTemplateListener.php
```diff
@@ -172,8 +172,9 @@ protected function deriveControllerClass($controller)
             $controller = substr($controller, strrpos($controller, '\\') + 1);
         }
 
-        if ((10 < strlen($controller))
-            && ('Controller' == substr($controller, -10))
+        $length = strlen($controller);
+        if ($length > 10
+            && strrpos($controller, 'Controller') === ($length - 10)
```
michalbundyra: Brackets are useless around `($length - 10)`.

michalbundyra (commented):
Not convinced to this change, imo it decreased readability.

src/View/Http/InjectTemplateListener.php
samsonasik: done ;)

Xerkus:
I would agree with @webimpress, this PR in its final form does not bring notable improvement. I am going to close this now.

Issue #185, opened 2016-07-14
[ZF3] zend-serializer is required


matbech:
In Controller\LazyControllerAbstractFacatory.php there is a reference to zend-serializer:
use Zend\Serializer\AdapterPluginManager as SerializerAdapterManager;

Therefore the zend-serializer is required in the composer.json dependencies or at minimum in the 'suggest' section.

Another option would be to remove the SerializerAdapterManager alias:
        SerializerAdapterManager::class => 'SerializerAdapterManager',
or replace the SerializerAdapterManager::class with the actual class name string.

weierophinney:
The lazy controller factory is opt-in. At most, we need to add a suggestion to the composer definition. Would you like to create a pull request with this change?

matbech:
LazyControllerAbstractFacatory uses additional aliases for which further suggestions (zend-console, zend-log, etc) would be required:
use Zend\Console\Adapter\AdapterInterface as ConsoleAdapterInterface;
use Zend\Log\FilterPluginManager as LogFilterManager;

```
    LogProcessorManager::class      => 'LogProcessorManager',
    LogWriterManager::class         => 'LogWriterManager',
```

So I'm not sure whether adding them all to the 'suggest' section is an elegant solution?

PR #281, opened 2018-05-16
#185 Add zend-serializer to composer suggestion.

panvid:


froschdesign (approved):

Xerkus:
I am not sure about this. It does not add any functionality to zend-mvc.

froschdesign:
@Xerkus 
If an end-user wants to use the `LazyControllerAbstractFactory` in his application, then zend-serializer is required – but only there.
Therefore the hint / suggestion in the Composer file for this _soft_ dependency.

https://github.com/zendframework/zend-mvc/blob/fe43713dd6d251343139142ad680eab049edb6c6/src/Controller/LazyControllerAbstractFactory.php#L21
---
At the moment we have no reference or guide for the Composer file, but we use the suggest section since the introduction of Composer.

Xerkus:
Neither `use` statement or magic `::class` constant require presence or existence of the referenced class. It is just a glorified string literal. I would argue this use case does not create soft dependency.

Controller requiring serializer adapter plugin manager as dependency would need to declare zend-serializer as a dependency on the other hand.

froschdesign:
@Xerkus 
You are right! The factory doesn't use the serializer for running.

PR #284, opened 2018-06-14
Minor comment update

bitwombat:
Clarifying the language used in `injectEventManager`'s comments.

The original statement that a shared event manager is injected into an existing event manager is incorrect - the code replaces the existing event manager with a new one.

I may have added too much 'how it works' detail, but I puzzled over this a bit so I thought the explanation might help future readers.


bitwombat:
Also, I wonder about the existing comment that "the conditional will always pass".  I think that's true.  Then does it need to be a conditional?

Issue #186, opened 2016-07-16
Class names for ServiceListenerFactory


RalfEggert:
I noticed that most services defined in the `Zend\Mvc\Service\ServiceListenerFactory` still use strings for the services names instead of class names. 

https://github.com/zendframework/zend-mvc/blob/develop/src/Service/ServiceListenerFactory.php

It looks a little inconsequent in a factory when all services are requested by class names except the very important `ViewHelperManager` for example. Is there a chance to get this sorted?

Example:

```
namespace UserFrontend\Authorization;

use Interop\Container\ContainerInterface;
use UserModel\Permissions\UserAcl;
use Zend\Authentication\AuthenticationService;
use Zend\ServiceManager\Factory\FactoryInterface;
use Zend\View\Helper\Navigation;

class AuthorizationListenerFactory implements FactoryInterface
{
    public function __invoke(
        ContainerInterface $container,
        $requestedName,
        array $options = null
    ) {
        $viewHelperManager = $container->get('ViewHelperManager');

        $authService      = $container->get(AuthenticationService::class);
        $userAcl          = $container->get(UserAcl::class);
        $navigationHelper = $viewHelperManager->get(Navigation::class);

        $authorizationListener = new AuthorizationListener(
            $authService, $userAcl, $navigationHelper
        );

        return $authorizationListener;
    }
}
```

froschdesign:
See at #259 

https://github.com/zendframework/zend-mvc/blob/bf93aa5e64e688a4f2bc7a50679c152bac289cae/src/ConfigProvider.php#L119

weierophinney:
This repository has been closed and moved to laminas/laminas-mvc; a new issue has been opened at https://github.com/laminas/laminas-mvc/issues/23.

PR #292, opened 2018-08-11
Make use of class names instead of strings for services

thexpand:
Related to issue: https://github.com/zendframework/zend-mvc/issues/186


thexpand:
The Travis is failing on master, too. What can be done?

geerteltink:
@thexpand Travis is failing on all `latest` builds, meaning it tests against all of the latest dependencies. So you could probably run `composer update` locally and should see the tests fail as well. It makes it easier to find the package that causes the error and fix it. And then you can still decide if you want to include the updated composer file or not.

thexpand:
@xtreamwayz I have created a hotfix for this as a new PR https://github.com/zendframework/zend-mvc/pull/293
We'll be able to continue with the current PR only after merging the hotfix. Can you make a code review?

boesing:
As already wrote in #294, we should close this PR in favor of #294.
Fixes required for passing unit tests are provided with #295.

If we close this, we can drop a lot of confusion with those Pull Requests I guess.

PR #290, opened 2018-07-18
add requirement of zend-stratigility to documentation

rieschl:
I recently updated a project from `zend-mvc` version 2.7 to 3.1 and ran into this issue. So I thought to add it to the documentation 🙂

PR #293, opened 2018-09-11
Provide support for zend-stratigility 2.2.2

thexpand:
Travis was failing on `latest` builds on all platforms. I've ran `composer update` and came across a deprecation error when I ran the unit tests. Callable double-pass middleware must be decorated in order to work with the new version of `zend-stratigility`.
That's what this hotfix is all about - it includes the updated Composer files and the decorated middleware.

### DIFF
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -16,7 +16,6 @@ use Zend\EventManager\EventManagerInterface;
 use Zend\Mvc\Controller\MiddlewareController;
 use Zend\Mvc\Exception\InvalidMiddlewareException;
 use Zend\Psr7Bridge\Psr7Response;
-use Zend\Stratigility\Middleware\DoublePassMiddlewareDecorator;
 use Zend\Stratigility\MiddlewarePipe;
 
 class MiddlewareListener extends AbstractListenerAggregate
@@ -141,16 +140,6 @@ class MiddlewareListener extends AbstractListenerAggregate
             }
             if (! $middlewareToBePiped instanceof MiddlewareInterface && ! is_callable($middlewareToBePiped)) {
                 throw InvalidMiddlewareException::fromMiddlewareName($middlewareName);
-            }
-
-            // Decorate double-pass middleware
-            if (is_callable($middlewareToBePiped)
-                && (
-                    ! $middlewaresToBePiped instanceof MiddlewareInterface
-                    || ! method_exists($middlewaresToBePiped, 'process')
-                )
-            ) {
-                $middlewareToBePiped = new DoublePassMiddlewareDecorator($middlewareToBePiped);
             }
 
             $pipe->pipe($middlewareToBePiped);
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -10,7 +10,6 @@ namespace Zend\Mvc;
 use Interop\Container\ContainerInterface;
 use Interop\Http\ServerMiddleware\MiddlewareInterface;
 use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface;
 use Zend\Mvc\Controller\MiddlewareController;
@@ -103,7 +102,7 @@ class MiddlewareListener extends AbstractListenerAggregate
 
         $event->setError('');
 
-        if (! $return instanceof PsrResponseInterface) {
+        if (! $return instanceof ResponseInterface) {
             $event->setResult($return);
             return $return;
         }
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -16,6 +16,7 @@ use Zend\EventManager\EventManagerInterface;
 use Zend\Mvc\Controller\MiddlewareController;
 use Zend\Mvc\Exception\InvalidMiddlewareException;
 use Zend\Psr7Bridge\Psr7Response;
+use Zend\Stratigility\Middleware\DoublePassMiddlewareDecorator;
 use Zend\Stratigility\MiddlewarePipe;
 
 class MiddlewareListener extends AbstractListenerAggregate
@@ -140,6 +141,16 @@ class MiddlewareListener extends AbstractListenerAggregate
             }
             if (! $middlewareToBePiped instanceof MiddlewareInterface && ! is_callable($middlewareToBePiped)) {
                 throw InvalidMiddlewareException::fromMiddlewareName($middlewareName);
+            }
+
+            // Decorate double-pass middleware
+            if (is_callable($middlewareToBePiped)
+                && (
+                    ! $middlewaresToBePiped instanceof MiddlewareInterface
+                    || ! method_exists($middlewaresToBePiped, 'process')
+                )
+            ) {
+                $middlewareToBePiped = new DoublePassMiddlewareDecorator($middlewareToBePiped);
             }
 
             $pipe->pipe($middlewareToBePiped);
--- a/src/MiddlewareListener.php
+++ b/src/MiddlewareListener.php
@@ -9,17 +9,13 @@ namespace Zend\Mvc;
 
 use Interop\Container\ContainerInterface;
 use Interop\Http\ServerMiddleware\MiddlewareInterface;
-use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ServerRequestInterface as PsrServerRequestInterface;
+use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Zend\EventManager\AbstractListenerAggregate;
 use Zend\EventManager\EventManagerInterface;
-use Zend\Mvc\Exception\InvalidMiddlewareException;
-use Zend\Mvc\Exception\ReachedFinalHandlerException;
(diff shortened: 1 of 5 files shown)


src/MiddlewareListener.php
```diff
@@ -1,25 +1,26 @@
 <?php
 /**
- * @link      http://github.com/zendframework/zend-mvc for the canonical source repository
- * @copyright Copyright (c) 2005-2018 Zend Technologies USA Inc. (http://www.zend.com)
+ * @see       https://github.com/zendframework/zend-mvc for the canonical source repository
```
froschdesign: Please use a separate commit for this changes. Don't mix different topics in one commit.

src/MiddlewareListener.php
```diff
@@ -1,25 +1,26 @@
...
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (http://www.zend.com)
```
froschdesign: Please revert this. (The current year is only needed for _new_ files.)

src/MiddlewareListener.php
```diff
@@ -146,6 +147,19 @@ private function createPipeFromSpec(
                 throw InvalidMiddlewareException::fromMiddlewareName($middlewareName);
             }
 
+            // Decorate double-pass middleware
+            if (is_callable($middlewareToBePiped)) {
+                $r = $this->getReflectionFunction($middlewareToBePiped);
```
froschdesign: Please use always meaningful names for variables. (loops are an exception)

But do we really need a reflection here? Smells like a hack.
A check of the interfaces and a simple `method_exists` is maybe enough.

froschdesign (changes requested):

src/MiddlewareListener.php
thexpand: This has been taken from `zend-stratigility` as a way of finding whether the middleware is a double-pass one, but I'll change it (it seemed like an over-complication to me, too).

froschdesign: > This has been taken from zend-stratigility…

Ah, good to know.

> it seemed like an over-complication to me, too

If the interface `Interop\Http\ServerMiddleware\MiddlewareInterface` is not present but the current middleware is callable, then there is also no check for the parameters.
And the decorator `DoublePassMiddlewareDecorator` includes also a test of the result.

froschdesign:
Related to #290

src/MiddlewareListener.php
thexpand: I've separated the commits and force-pushed them.

src/MiddlewareListener.php
thexpand: @froschdesign 
I didn't know that. Can you confirm the following, just to be clear for future updates?
1. If a file is edited and the ending year of the range is not the current year, it must be updated (e.g. 2005-2017 becomes 2005-2018)
2. If a file is edited and there is no ending year, the current year must be added at the end (e.g. 2017 becomes 2017-2018)
3. For new files the current year must be added (2018)

src/MiddlewareListener.php
thexpand: I've applied changes. Now we check whether the callable is missing the `process` method or if it does not implement the `Interop\Http\ServerMiddleware\MiddlewareInterface`, then wraps it, using the decorator.

src/MiddlewareListener.php
froschdesign: @thexpand 
1., 2. and 3. are correct! 👍

src/MiddlewareListener.php
```diff
@@ -146,6 +143,16 @@ private function createPipeFromSpec(
                 throw InvalidMiddlewareException::fromMiddlewareName($middlewareName);
             }
 
+            // Decorate double-pass middleware
+            if (is_callable($middlewareToBePiped)
+                && (
+                    ! $middlewaresToBePiped instanceof MiddlewareInterface
+                    || ! method_exists($middlewaresToBePiped, 'process')
```
froschdesign: These changes should covered by unit tests.

froschdesign (changes requested):

src/MiddlewareListener.php
thexpand: A new test `testSuccessfullyDispatchesDoublePassMiddleware`?

froschdesign: For example.

thexpand: I've added the test, but as I come to think about the if condition, doesn't it have to be `x && !y && !z` instead of `x && (!y || !z)`.. the first one makes more sense than the second one. Otherwise middlewares that are not an instance of `MiddlewareInterface`, but have the `process` method would be decorated, as well. Hm...

src/MiddlewareListener.php
```diff
@@ -9,17 +9,14 @@
 
 use Interop\Container\ContainerInterface;
 use Interop\Http\ServerMiddleware\MiddlewareInterface;
-use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ServerRequestInterface as PsrServerRequestInterface;
+use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
```
geerteltink: `Psr\Http\Message\ResponseInterface` is imported twice?

geerteltink (commented):
I've been digging a bit deeper into the failed tests and this PR. What you are basically doing is decorating the middleware in the tests to prevent the deprecation notice from stratigility. Doing it this way it will prevent notices for users as well. By the time zend-mvc is updated to support stratigility 3.0 their applications are not working anymore and they never saw a notice.

I think the correct way is to actually fix the middleware listener tests.
https://github.com/zendframework/zend-mvc/blob/a8d45689d37a9e4ff4b75ea0b7478fa3d4f9c089/test/MiddlewareListenerTest.php#L62-L69
As you can see that test uses the double pass middleware. This should be changed to use the PSR-12 Middleware interface:
`public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface;`

src/MiddlewareListener.php
thexpand: It was there before the changes, but I fixed it anyway.

geerteltink:
To summarize the discussion we had last night on Slack:
- Currently all zend-mvc `latest` tests fail because of a deprecation notice in stratigility 2.2. The notices are there for a reason and shouldn't be suppressed. See [here](https://travis-ci.org/zendframework/zend-mvc/builds/427665383) and [here](https://travis-ci.org/zendframework/zend-mvc/builds/415071845).
- Stratigility 3.0 uses PSR-15
- Stratigility 2.2 has the `DoublePassMiddlewareDecorator` which is needed to properly suppress the deprecation notice.
- Stratigility 2.1 has support for http-interop/http-middleware 0.5.0 via a polyfill.
- Stratigility 2.0 has support for http-interop/http-middleware version to 0.4.1

zend-mvc is tested against stratigility ^2.0.1. The deprecation notices were added in 2.2.0.

Since stratigility 3 requires php ^7.1 and zend-mvc requires ^5.6 || ^7.0, support for PSR-15 is out of the question. We tried to make expressive 2 compatible with all different middleware interfaces and it became pretty messy.

I suggested this solution:
1. Release a zend-mvc `2.7.16` hotfix that restricts stratigility in composer require-dev to `"zendframework/zend-stratigility": ">=2.0.1 <2.2"` and fix the tests.
2. Release zend-mvc `2.8.0` with support for `"zendframework/zend-stratigility": "^2.2.2"` and use the `DoublePassMiddlewareDecorator` in the `MiddlewareListenerTest` to fix the current failing `latest` tests.

/cc @weierophinney

froschdesign:
@xtreamwayz 
Sorry, maybe I missed something. But why 2.7. and 2.8?

thexpand:
The `master` branch is currently on `3.1.1`, so the hotfix will be released as a new patch version `3.1.2` and then the update to support stratigility `2.2.2` will be released as a new minor version `3.2.0`.

weierophinney:
> I suggested this solution:
> 
>     1. Release a zend-mvc `2.7.16` hotfix that restricts stratigility in composer require-dev to `"zendframework/zend-stratigility": ">=2.0.1 <2.2"` and fix the tests.
> 
>     2. Release zend-mvc `2.8.0` with support for `"zendframework/zend-stratigility": "^2.2.2"` and use the `DoublePassMiddlewareDecorator` in the `MiddlewareListenerTest` to fix the current failing `latest` tests.

The v2 series of zend-mvc is no longer supported (see the [LTS page for details](https://framework.zend.com/long-term-support)). As such, we can ignore those versions entirely. Users will have to heed the notices and update their code accordingly.

For v3, the approach @thexpand outlines seems sensible.

geerteltink:
> Sorry, maybe I missed something. But why 2.7. and 2.8?

No, I missed something... the zend-mvc 3 release :D

boesing:
So I guess this can be closed in favor of #295?

thexpand:
@boesing This PR has nothing to do with #295 
It is about providing support for stratigility `2.2.2`.
As outlined above, this is the second step. The PR you are talking about is meant to _only_ fix the failing tests.

boesing:
@thexpand ah never mind, just saw the stratigility change and thought this was the the same composer change as in #295
So just ignore my comment.

Xerkus:
I marked this PR as won't fix but will keep it open for now. Next week I will pull the stratigility support into satellite mvc package and reapply 2.2.2 changes there.

Xerkus:
Alternative solution to be provided by #308

PR #296, opened 2018-10-13
PHPStan fixes

thomasvargiu:
In order to bump to PHP 7 minimum version in the future, I used phpstan to static analyse the code, correcting some phpdoc errors and adding guards throwing exceptions when needed. Code quality will be improved when other components will be improved.

It's just a first step, but I think it will be useful.

I don't know if this could be considered a BC break, probably it isn't.


src/Controller/AbstractRestfulController.php
```diff
@@ -210,13 +227,13 @@ public function options()
      * Not marked as abstract, as that would introduce a BC break
      * (introduced in 2.1.0); instead, raises an exception if not implemented.
      *
-     * @param  $id
-     * @param  $data
+     * @param mixed $id
```
Ocramius, line 230: These are `string`, `mixed`

src/Controller/AbstractRestfulController.php
```diff
@@ -332,6 +349,20 @@ public function onDispatch(MvcEvent $e)
 
         $request = $e->getRequest();
 
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\RuntimeException(
+                'Request is not an instance of ' . HttpRequest::class
+            );
+        }
+
+        $response = $e->getResponse();
```
Ocramius: Group the two assignments together, then have the two check blocks (easier to read)

src/Controller/AbstractRestfulController.php
```diff
@@ -546,11 +587,11 @@ public function addHttpMethodHandler($method, /* Callable */ $handler)
...
      * query string, returning it if found. Otherwise, returns a boolean false.
      *
-     * @param  \Zend\Router\RouteMatch $routeMatch
-     * @param  Request $request
+     * @param  RouteMatch $routeMatch
+     * @param  HttpRequest $request
      * @return false|mixed
      */
-    protected function getIdentifier($routeMatch, $request)
+    protected function getIdentifier(RouteMatch $routeMatch, HttpRequest $request)
```
Ocramius: This is a BC break that needs to be reversed

src/Controller/Plugin/AcceptableViewModelSelector.php
```diff
@@ -86,14 +88,22 @@ public function getViewModel(
...
         if (! class_exists($name)) {
             throw new InvalidArgumentException('The supplied View Model could not be found');
         }
 
-        return new $name();
+        $viewModel = new $name();
+
+        if (! $viewModel instanceof ModelInterface) {
+            throw new InvalidArgumentException(
+                'The supplied View Model is not an instance of ' . ModelInterface::class
```
Ocramius: Please add `$name` to the exception message

src/Controller/AbstractRestfulController.php
```diff
@@ -90,6 +93,20 @@ public function getIdentifierName()
...
 
+    /**
+     * @return HttpResponse
+     */
+    private function getHttpResponse()
+    {
+        $response = $this->getResponse();
+        
+        if (! $response instanceof HttpResponse) {
+            throw new Exception\RuntimeException('Response is not an instance of ' . HttpResponse::class);
```
Ocramius: Please add the type of `$response` to this exception

Ocramius: Possibly an `UnexpectedValueException` btw - not a `RuntimeException`.

src/Controller/AbstractRestfulController.php
```diff
@@ -332,6 +349,20 @@ public function onDispatch(MvcEvent $e)
 
         $request = $e->getRequest();
 
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\RuntimeException(
+                'Request is not an instance of ' . HttpRequest::class
```
Ocramius: Type of the found value is to be added to the exception

src/Controller/AbstractRestfulController.php
```diff
@@ -332,6 +349,20 @@ public function onDispatch(MvcEvent $e)
...
+            );
+        }
+
+        $response = $e->getResponse();
+
+        if (! $response instanceof HttpResponse) {
+            throw new Exception\RuntimeException(
+                'Request is not an instance of ' . HttpRequest::class
```
Ocramius: Type of the found value is to be added to the exception

src/Controller/AbstractRestfulController.php
```diff
@@ -462,6 +491,12 @@ public function onDispatch(MvcEvent $e)
      */
     public function processPostData(Request $request)
     {
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\InvalidArgumentException(
+                'Request is not an instance of ' . HttpRequest::class
```
Ocramius: Type of the found value is to be added to the exception

src/Controller/AbstractRestfulController.php
```diff
@@ -478,7 +513,13 @@ public function processPostData(Request $request)
      */
     public function requestHasContentType(Request $request, $contentType = '')
     {
-        /** @var $headerContentType \Zend\Http\Header\ContentType */
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\InvalidArgumentException(
+                'Request is not an instance of ' . HttpRequest::class
```
Ocramius: Type of the found value is to be added to the exception

src/Controller/Plugin/Forward.php
```diff
@@ -125,7 +126,13 @@ public function dispatch($name, array $params = null)
         // Allow passing parameters to seed the RouteMatch with & copy matched route name
         if ($params !== null) {
             $routeMatch = new RouteMatch($params);
-            $routeMatch->setMatchedRouteName($event->getRouteMatch()->getMatchedRouteName());
+            $eventRouteMatch = $event->getRouteMatch();
+
+            if (! $eventRouteMatch) {
```
Ocramius: You previously used `null !==`, while here you used `!`: pick one of the two approaches please :)

src/View/Http/RouteNotFoundStrategy.php
```diff
@@ -147,6 +148,13 @@ public function detectNotFoundError(MvcEvent $e)
                     $response = new HttpResponse();
                     $e->setResponse($response);
                 }
+
+                if (! $response instanceof HttpResponse) {
+                    throw new RuntimeException(
+                        'Event response should be an instance of ' . HttpResponse::class
```
Ocramius: Please add the type of `$response` here

src/View/Http/RouteNotFoundStrategy.php
```diff
@@ -169,6 +177,11 @@ public function prepareNotFoundViewModel(MvcEvent $e)
         }
 
         $response = $e->getResponse();
+
+        if (! $response instanceof HttpResponse) {
+            throw new RuntimeException('Event response is not an instance of ' . HttpResponse::class);
```
Ocramius: Please add the type of `$response` here

src/View/Http/InjectTemplateListener.php
```diff
@@ -75,6 +81,10 @@ public function injectTemplate(MvcEvent $e)
 
         $template = $this->mapController($controller);
 
+        if (false === $template) {
+            throw new RuntimeException('Invalid controller name');
```
Ocramius: Please add the name of the controller

src/View/Http/ExceptionStrategy.php
```diff
@@ -129,11 +130,13 @@ public function prepareExceptionViewModel(MvcEvent $e)
...
                     $response->setStatusCode(500);
                     $e->setResponse($response);
-                } else {
+                } elseif ($response instanceof HttpResponse) {
                     $statusCode = $response->getStatusCode();
                     if ($statusCode === 200) {
                         $response->setStatusCode(500);
                     }
+                } else {
+                    throw new RuntimeException('Event response is not an instance of ' . HttpResponse::class);
```
Ocramius: Please add the type of `$response` here

src/Service/ApplicationFactory.php
```diff
@@ -26,6 +28,10 @@ class ApplicationFactory implements FactoryInterface
      */
     public function __invoke(ContainerInterface $container, $name, array $options = null)
     {
+        if (! $container instanceof ServiceManager) {
+            throw new InvalidArgumentException('Container should be an instance of ' . ServiceManager::class);
```
Ocramius: Please add the type of `$container` here

src/ResponseSender/SimpleStreamResponseSender.php
```diff
@@ -23,9 +24,16 @@ public function sendStream(SendResponseEvent $event)
             return $this;
         }
         $response = $event->getResponse();
+
+        if (! $response instanceof Stream) {
+            throw new RuntimeException('Event is not an instance of ' . Stream::class);
```
Ocramius: Please add the type of `$response` here

Also: `Stream`?

src/ResponseSender/AbstractResponseSender.php
```diff
@@ -25,11 +28,16 @@ public function sendHeaders(SendResponseEvent $event)
 
         $response = $event->getResponse();
 
+        if (! $response instanceof Response) {
+            throw new RuntimeException('Event response should be an instance of ' . Response::class);
```
Ocramius: Please add the type of `$response` here

Ocramius (changes requested):
Overall good improvement, but:

 1. there's a lot of constructor code landing in here, making things complex to understand/follow. Strongly endorse using `ExceptionType::fromSomething($parameters)` instead of the default `Exception::__construct()` API, which is quite fragile
 2. can we add phpstan to a build step? Would need to be locked to a specific version of phpstan

src/Controller/AbstractRestfulController.php
thomasvargiu: On other methods `$id` is `mixed`. Should we change the `$id` type on the other methods too?

Ocramius: No, let's keep it as `mixed` then: I assume the ID is parsed and is the responsibility of the subclass then

src/ResponseSender/SimpleStreamResponseSender.php
thomasvargiu: Yes, it's the only response type with `getStream()` method.

There is a chain to send the response, the first one is `PhpEnvironmentResponseSender` which has a method to check if the event response is an instance of `Zend\Http\PhpEnvironment\Response`.

I added the check also for the `SimpleStreamResponseSender`.

src/View/Http/InjectTemplateListener.php
thomasvargiu: I'm going to change the phpdoc because `mapController()` cannot return a `false` value as phpdoc says.

thomasvargiu:
@Ocramius I can add phpstan on CI, but just a low level. Too many errors depends on other libraries.

Ocramius:
@thomasvargiu that's fine - we can raise the level as we get better

Ocramius:
@thomasvargiu btw, I still suggest getting rid of all the `throw new SomeException($someString)`, and having `throw SomeException::unexpectedType($expected, $actual)` instead

thomasvargiu:
@Ocramius I'm still working on this. I set this PR as WIP

thomasvargiu:
Done. Needs a complete review

phpstan.neon
```diff
@@ -0,0 +1,15 @@
+parameters:
```
Ocramius: Please add the phpstan level and the file locations here, so it picks them up by default.

composer.json
```diff
@@ -30,8 +31,15 @@
         "http-interop/http-middleware": "^0.4.1",
         "phpunit/phpunit": "^6.4.4 || ^5.7.14",
         "zendframework/zend-coding-standard": "~1.0.0",
+        "zendframework/zend-filter": "^2.6",
```
Ocramius: That's a helluvalot of added dependencies that may as well fail in production too :|

src/Controller/AbstractRestfulController.php
```diff
@@ -462,6 +486,14 @@ public function onDispatch(MvcEvent $e)
      */
     public function processPostData(Request $request)
     {
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/Controller/AbstractRestfulController.php
```diff
@@ -473,20 +505,31 @@ public function processPostData(Request $request)
...
-     * @param  string|null $contentType
+     * @param  string $contentType
      * @return bool
      */
     public function requestHasContentType(Request $request, $contentType = '')
     {
-        /** @var $headerContentType \Zend\Http\Header\ContentType */
-        $headerContentType = $request->getHeaders()->get('content-type');
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/Controller/AbstractRestfulController.php
```diff
@@ -546,12 +589,28 @@ public function addHttpMethodHandler($method, /* Callable */ $handler)
...
-     * @param  \Zend\Router\RouteMatch $routeMatch
-     * @param  Request $request
+     * @param  RouteMatch $routeMatch
+     * @param  HttpRequest $request
      * @return false|mixed
      */
     protected function getIdentifier($routeMatch, $request)
     {
+        if (! $routeMatch instanceof RouteMatch) {
+            throw new Exception\InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/Controller/AbstractRestfulController.php
```diff
@@ -546,12 +589,28 @@ public function addHttpMethodHandler($method, /* Callable */ $handler)
...
+                'RouteMatch must be an instance of %s. %s given',
+                RouteMatch::class,
+                \is_object($routeMatch) ? \get_class($routeMatch) : \gettype($routeMatch)
+            ));
+        }
+
+        if (! $request instanceof HttpRequest) {
+            throw new Exception\InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/Controller/Plugin/AcceptableViewModelSelector.php
```diff
@@ -86,14 +90,25 @@ public function getViewModel(
...
 
         if (! class_exists($name)) {
             throw new InvalidArgumentException('The supplied View Model could not be found');
         }
 
-        return new $name();
+        $viewModel = new $name();
+
+        if (! $viewModel instanceof ModelInterface) {
+            throw new InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/Controller/Plugin/AcceptableViewModelSelector.php
```diff
@@ -220,7 +242,11 @@ protected function injectViewModelName($modelAcceptString, $modelName)
      */
     protected function extractViewModelName(AbstractFieldValuePart $res)
     {
-        $modelName = $res->getMatchedAgainst()->params[self::INJECT_VIEWMODEL_NAME];
+        $matchedAgainst = $res->getMatchedAgainst();
+        if (null === $matchedAgainst) {
+            throw new UnexpectedValueException('Unable to find matchedAgainst value');
+        }
+        $modelName = $matchedAgainst->params[self::INJECT_VIEWMODEL_NAME];
```
Ocramius, line 244: `$modelName` is a redundant variable

src/Controller/Plugin/Params.php
```diff
@@ -7,11 +7,24 @@
...
+use Zend\Mvc\MvcEvent;
 
 class Params extends AbstractPlugin
 {
+    private function getPluginController()
+    {
+        $controller = $this->getController();
+
+        if (! $controller instanceof AbstractController) {
+            throw new RuntimeException('Controller is not an instance of ' . AbstractController::class);
```
Ocramius, line 22: Same pattern about exception constructor to be applied here as well

src/Controller/Plugin/Params.php
```diff
@@ -101,18 +114,30 @@ public function fromQuery($param = null, $default = null)
...
         if (! $controller instanceof InjectApplicationEventInterface) {
             throw new RuntimeException(
                 'Controllers must implement Zend\Mvc\InjectApplicationEventInterface to use this plugin.'
             );
         }
 
+        $event = $controller->getEvent();
+
+        if (! $event instanceof MvcEvent) {
+            throw new RuntimeException('Controller event is not an instance of ' . MvcEvent::class);
```
Ocramius, line 128: Same pattern about exception constructor to be applied here as well

src/DispatchListener.php
```diff
@@ -75,22 +76,24 @@ public function attach(EventManagerInterface $events, $priority = 1)
...
         $routeMatch        = $e->getRouteMatch();
         $controllerName    = $routeMatch instanceof RouteMatch
             ? $routeMatch->getParam('controller', 'not-found')
             : 'not-found';
         $application       = $e->getApplication();
-        $events            = $application->getEventManager();
         $controllerManager = $this->controllerManager;
 
+        if (! $application instanceof Application) {
+            throw new RuntimeException('Application is not an instance of ' . Application::class);
```
Ocramius, line 90: Same pattern about exception constructor to be applied here as well

src/MiddlewareListener.php
```diff
@@ -44,20 +44,38 @@ public function attach(EventManagerInterface $events, $priority = 1)
...
 
         $request        = $event->getRequest();
         $application    = $event->getApplication();
+
+        if (null === $request) {
+            throw new UnexpectedValueException('No Request in event');
+        }
+
+        if (! $application instanceof Application) {
+            throw new RuntimeException('Application is not an instance of ' . Application::class);
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/MiddlewareListener.php
```diff
@@ -44,20 +44,38 @@ public function attach(EventManagerInterface $events, $priority = 1)
...
+        }
+
         $response       = $application->getResponse();
         $serviceManager = $application->getServiceManager();
 
+        if (! $response instanceof Response) {
+            throw new RuntimeException('Application response is not an instance of ' . Response::class);
```
Ocramius: Same pattern about exception constructor to be applied here as well

src/MiddlewareListener.php
```diff
@@ -44,20 +44,38 @@ public function attach(EventManagerInterface $events, $priority = 1)
...
         }
 
         $request        = $event->getRequest();
         $application    = $event->getApplication();
+
+        if (null === $request) {
+            throw new UnexpectedValueException('No Request in event');
+        }
+
+        if (! $application instanceof Application) {
```
Ocramius, line 68: `ApplicationInterface`

src/MiddlewareListener.php
Ocramius: `ApplicationInterface::class`

src/Service/ApplicationFactory.php
```diff
@@ -26,6 +28,14 @@ class ApplicationFactory implements FactoryInterface
      */
     public function __invoke(ContainerInterface $container, $name, array $options = null)
     {
+        if (! $container instanceof ServiceManager) {
+            throw new InvalidArgumentException(sprintf(
```
Ocramius: Same pattern about exceptions

src/Service/HttpViewManagerConfigTrait.php
```diff
@@ -22,11 +22,15 @@ private function getConfig(ContainerInterface $container)
...
-        if (isset($config['view_manager'])
-            && (is_array($config['view_manager'])
-                || $config['view_manager'] instanceof ArrayAccess
-            )
-        ) {
+        if (! isset($config['view_manager'])) {
+            return [];
+        }
+
+        if ($config['view_manager'] instanceof ArrayAccess) {
```
Ocramius: Careful: an `(array)` cast of an object is potentially destructive. Prefer asserting that it `is_array()` instead

src/View/Http/ExceptionStrategy.php
```diff
@@ -129,11 +131,13 @@ public function prepareExceptionViewModel(MvcEvent $e)
                     $response = new HttpResponse();
                     $response->setStatusCode(500);
                     $e->setResponse($response);
-                } else {
+                } elseif ($response instanceof HttpResponse) {
                     $statusCode = $response->getStatusCode();
                     if ($statusCode === 200) {
                         $response->setStatusCode(500);
                     }
+                } else {
```
Ocramius: Instead of using an `else`, put early `break;` statements above

Ocramius (changes requested):
Good improvement, and almost there :+1: 

Testing the exception class, although it may sound silly, is also something I'd really want here.

composer.json
thomasvargiu: There are classes used in `LazyControllerAbstractFactory`. The static analyser throws an exception if they are not found. We can convert the `::class` notation in strings, probably is the best thing to do.

Btw, they are included only in `require-dev` directive so shouldn't be a problem in production.

Ocramius: If they are in `::class` form, then using the string notation is preferable over adding dependencies here

src/MiddlewareListener.php
thomasvargiu: There is a method (`marshalInvalidMiddleware`) called some lines lower, where the signature require an `Application` instance. The interface would be enough but to avoid a BC break I can't change the signature.

src/Service/HttpViewManagerConfigTrait.php
thomasvargiu: Thre is a problem with the `config` service in `zend-mvc`. Somewhere is declared as `array|ArrayAccess`, but then is passed on methods that accept just an `array`. This will be a problem with strict types.

I would prefer to return just an `array`. You're right, it's dangerous. What about to convert it iterating the array?

Ocramius: That's perfectly valid, but an `(array)` cast produces something quite broken: https://3v4l.org/OTFXt

```php
<?php

class A implements ArrayAccess {
    private $potato = 'haha';
    public function offsetExists($offset){}
    public function offsetGet($offset){}
    public function offsetSet ($offset, $value){}
    public function offsetUnset ($offset){}
}

var_dump((array) new A());
```

thomasvargiu: I know. Do you agree that should be better to return just an array? In order to use strict types in the future I think should be better IMHO.

I think it's not possible to convert safely an object implementing `ArrayAccess` in `array`, so I should revert it, but it's a problem to handle it in other classes

thomasvargiu: The return value was used only by private methods, so I changed the signature to allow `array|ArrayAccess`

weierophinney:
This repository has been closed and moved to laminas/laminas-mvc; a new issue has been opened at https://github.com/laminas/laminas-mvc/issues/9.

weierophinney:

PR #301, opened 2019-02-03
Bump minimum php to 7.2 for next major

Xerkus:
Note: PHP 7.1 is out of active support and have only 9 months of security support left at this time.


.gitignore
```diff
@@ -1,7 +1,10 @@
+.phpunit.result.cache
```
michalbundyra, line 1: Why there is no slash at the beginning? Can it be in any place?

.gitignore
```diff
@@ -1,7 +1,10 @@
...
+/.phpcs-cache
 /clover.xml
 /coveralls-upload.json
 /docs/html/
 /phpunit.xml
 /vendor/
 /zf-mkdoc-theme.tgz
 /zf-mkdoc-theme/
+
```
michalbundyra: We can remove that empty line.

.travis.yml
```diff
@@ -11,52 +11,20 @@ env:
...
+        - TEST_COVERAGE=true
     - php: 7.2
       env:
         - DEPS=latest
     - php: 7.3
       env:
         - DEPS=lowest
-    - php: 7.3
-      env:
-        - DEPS=locked
```
michalbundyra: Why not testing locked on PHP 7.3?

composer.json
```diff
@@ -16,8 +16,8 @@
         "forum": "https://discourse.zendframework.com/c/questions/components"
     },
     "require": {
+        "php": "^7.2",
```
michalbundyra, line 19: ❤️

test/Service/ServiceListenerFactoryTest.php
```diff
@@ -7,42 +7,49 @@
...
         $this->factory  = new ServiceListenerFactory();
     }
 
-    /**
-     * @expectedException        Zend\ServiceManager\Exception\ServiceNotCreatedException
-     * @expectedExceptionMessage The value of service_listener_options must be an array, string given.
-     */
     public function testInvalidOptionType()
     {
+        $this->expectException(ServiceNotCreatedException::class);
```
michalbundyra, line 38: For me it will be much clearer to put these expectations just before the function which should throw the exception.
In that case it doesn't make a big difference, as the other line here is mocking, but in many case it could be wrong adding it at the very beginning part of the test.

michalbundyra (commented):
Is it just the bump (and PHPUnit upgrade) and other changes will be in separate PR? I would expect to use some PHP 7.2 features in the next major release.

.gitignore
Xerkus: Yes.
It was placed in folder next to a test when I run just that specific test from PhpStorm. Probably CWD was not a project root.

.gitignore
Xerkus: I think it was present in maintainers template.

.travis.yml
Xerkus: Locked dependencies are there to tell us when breakage comes from dependency and not the change in PR. Having lock file for a single version is sufficient for that, since it is made for specific version and it avoids errors like we had with Doctrine instantiator all over the framework because of lock file on wrong version.

test/Service/ServiceListenerFactoryTest.php
Xerkus: This exactly mirrors deprecated annotation behavior.
Whole test suite needs a few improvement passes. Not in this PR/at this time.

Xerkus:
> Is it just the bump (and PHPUnit upgrade) and other changes will be in separate PR? I would expect to use some PHP 7.2 features in the next major release.

It is just a baseline PR. Next will be coding standard upgrade, followed by typehints PR followed by actual major changes

.travis.yml
michalbundyra: Still we can have this issue, if someone regenerate new lock file on PHP 7.3. But I see your point. Just thinking maybe it will be better to run on the "latest" stable, supported PHP version then?

Xerkus: If packages are incompatible, PR will error out. A fix would be to update packages on different version. No legacy deps workarounds in build config.

Makes sense. Then cs check and coverage needs to be moved to latest too.

Ocramius (approved):
:ship:

